Issue #74, opened 2026-04-17
Ashim Docker image over 29GB


mariushosting:
### Before submitting

- [x] I searched existing issues and didn't find a duplicate
- [x] I'm running the latest version of ashim

### Description

How is it possible that this Docker image grows by 2–3 GB with every new version? Something doesn't add up here.

<img width="1881" height="733" alt="Image" src="https://github.com/user-attachments/assets/acbe2106-37fe-43f7-aa41-d165058a6b4c" />

### Steps to reproduce

https://hub.docker.com/r/ashimhq/ashim/tags

### Docker setup

```shell
Latest Docker
```

### Image tag

1.15.8

### Platform

AMD64 (Intel/AMD)

### Logs

```shell

```

Kofl:
Can confirm the docker size image.

ashim-hq:
Hey, thanks for reporting this - the image size is definitely a problem right now.

All the AI/ML models (background removal, upscaling, OCR, etc.) are currently bundled directly into the Docker image, which is what's inflating it to ~29 GB.

We're actively working on an on-demand approach where the base image ships with just the core features, and AI models can be downloaded individually after installation. This should bring the base image size down significantly while still giving you access to the full feature set when you need it.

**ETA: April 26th**

I'll update this issue once the release is out.

PR #78, opened 2026-04-18
feat: multi-GPU support, lazy model downloads, and Helm chart

### LINKED ISSUE #77: Images can not be pulled/ Image size 29 GB???
### Before submitting

- [x] I searched existing issues and didn't find a duplicate
- [x] I'm running the latest version of ashim

### Description

Latest Images can not be pulled, from non of the 2 mentioned sources.  Only with explicit tag: 1.15.9, not with latest
Image size 29 GB???

### Steps to reproduce

see above

### Docker setup

```shell
see above
```

### Image tag

latest

### Platform

AMD64 (Intel/AMD)

### Browser

_No response_

### Logs

```shell

```

### Screenshots

_No response_

swheettaos:
## Summary

Fixes #77, #74 — the 29 GB image that breaks CI and prevents `latest` from being updated.

---

## Root Cause

**Issue #77, #74: image size 29 GB, `latest` tag never updated**

The problem has two compounding parts:

**1. Base image bloat (~20 GB)**
`nvidia/cuda:12.6.3-cudnn-runtime-ubuntu24.04` ships the full CUDA runtime, cuDNN libraries, and all NVIDIA toolkit dependencies. This alone consumes ~20 GB of disk before a single line of application code is installed. GitHub-hosted `ubuntu-latest` runners have ~30 GB of usable disk after the pre-installed toolchain — leaving almost no headroom.

**2. Models baked in at build time (~5–6 GB)**
The original Dockerfile ran `download_models.py` during the image build, embedding ONNX and other model weights directly into the image layer. Combined with the CUDA base, the final image reached ~29 GB.

**3. The cascade failure**
When the amd64 build exhausts disk space mid-layer, `docker build` fails silently with a no-space error. The build job is marked failed, but because the manifest job only runs when *all* build matrix jobs succeed, it is skipped. The `latest` tag on both GHCR and Docker Hub is never updated — so users pulling `latest` get an increasingly stale image with no error or warning.

**4. Why the arm64 build appeared fine**
The arm64 variant used a plain `ubuntu` base (no CUDA — CUDA has no arm64/aarch64 support in this image family), so it was small enough to build successfully. This masked the problem: `latest` was technically being published from the arm64 manifest only, but amd64 users silently received the wrong architecture or a pull error depending on their client.

---

#### Docker image (`docker/Dockerfile`)
- Base changed from `nvidia/cuda` → `ubuntu:24.04` for all builds
- New `GPU_BACKEND=cpu|cuda|rocm` build arg controls which GPU pip wheels are installed
- New `BAKE_MODELS=false` build arg — models are **not** baked in by default
- ONNX runtime variant selected per backend (`onnxruntime` / `onnxruntime-gpu` / `onnxruntime-rocm`)
- Torch, PaddlePaddle, rembg all gated on `GPU_BACKEND`
- NCCL symlink fix only applied for `cuda` backend
- `download_models.py` kept permanently at `/opt/ashim/download_models.py`
- `latest-cuda` image bakes `NVIDIA_VISIBLE_DEVICES=all` + `NVIDIA_DRIVER_CAPABILITIES=compute,utility` via build-args so an NVIDIA runtime-enabled host injects GPU automatically (see GPU usage below)

#### First-run model download (`docker/entrypoint.sh`)
- On first container start, if `/opt/models/.complete` sentinel is absent, models are downloaded via `download_models.py` (runs as `ashim` user via gosu)
- Sentinel written after successful download; subsequent starts skip the download

#### docker-compose (`docker/docker-compose.yml`)
- Added `ashim-models:/opt/models` named volume for persistent model storage
- Added commented GPU override examples for NVIDIA and AMD

#### Release pipeline (`.github/workflows/release.yml`)
- Matrix expanded to 4 entries: `linux/amd64+cpu`, `linux/arm64+cpu`, `linux/amd64+cuda`, `linux/amd64+rocm`
- Three separate manifest groups: `latest` (cpu, multi-arch), `latest-cuda` (NVIDIA, amd64), `latest-rocm` (AMD, amd64)
- Manifest job now runs if *any* build succeeds (not all) — `latest` is always published as long as both cpu builds pass
- `GPU_BACKEND` build arg passed per matrix entry; NVIDIA env vars passed as build-args for the cuda entry only

#### CI (`.github/workflows/ci.yml`)
- Switched from `GHCR_TOKEN` secret → built-in `GITHUB_TOKEN` with `packages: write`
- Simplified registry cache (no conditional expressions)

#### Helm chart (`charts/ashim/`)
- New chart for Kubernetes deployment
- PVCs always created: `models` (10Gi), `data` (5Gi)
- Ingress disabled by default; GPU tolerations/nodeSelector configurable
- Published via GitHub Pages with `.github/workflows/chart-release.yml`

#### Documentation (`README.md`, `apps/docs/`)
- `README.md`: GPU quick start now shows `latest-cuda` + `--gpus all` for NVIDIA; adds AMD collapse section with `latest-rocm` + `--device` flags
- `apps/docs/guide/getting-started.md`: GPU tip updated to `latest-cuda`; AMD tip added for `latest-rocm`
- `apps/docs/guide/docker-tags.md`: rewritten to document the new 3-tag structure — new intro table, NVIDIA/AMD sections with correct commands, Docker Compose examples for all three variants (`latest`, `latest-cuda` with deploy section, `latest-rocm` with devices section), corrected platforms table (ARM64 always `latest`), versioned tag examples (`1.16.0-cuda` etc.), model download volume guidance, corrected migration note

---

## Image tags after this change

| Tag | Platform | GPU |
|---|---|---|
| `latest` | amd64 + arm64 | CPU only |
| `latest-cuda` | amd64 | NVIDIA (CUDA wheels baked in) |
| `latest-rocm` | amd64 | AMD (ROCm wheels baked in) |

---

#### NVIDIA

The `latest-cuda` image bakes `NVIDIA_VISIBLE_DEVICES=all` and `NVIDIA_DRIVER_CAPABILITIES=compute,utility` into its env layer. When the [NVIDIA Container Runtime](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/install-guide.html) is configured as the **default** Docker runtime on the host, these env vars trigger automatic GPU injection — no `--gpus all` flag needed:

```bash
docker run -p 1349:1349 -v ashim-models:/opt/models ghcr.io/ashim-hq/ashim:latest-cuda
```

If the NVIDIA runtime is *not* the default runtime, add `--gpus all` explicitly:

```bash
docker run --gpus all -p 1349:1349 -v ashim-models:/opt/models ghcr.io/ashim-hq/ashim:latest-cuda
```

#### AMD

ROCm has no equivalent auto-injection mechanism. Device passthrough is always required:

```bash
docker run \
  --device=/dev/kfd --device=/dev/dri \
  -p 1349:1349 -v ashim-models:/opt/models \
  ghcr.io/ashim-hq/ashim:latest-rocm
```

---

## Breaking changes

`latest` no longer includes CUDA. Users who previously ran `docker run --gpus all ... ashimhq/ashim:latest` for GPU acceleration should switch to `latest-cuda`. CPU users and ARM64 users are unaffected.

### DIFF
--- a/.github/workflows/chart-release.yml
+++ b/.github/workflows/chart-release.yml
@@ -1,0 +1,91 @@
+name: Release Helm Chart
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'charts/**'
+  workflow_dispatch:
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Configure Git
+        run: |
+          git config user.name "$GITHUB_ACTOR"
+          git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
+
+      - name: Install Helm
+        uses: azure/setup-helm@v4
+
+      - name: Bootstrap gh-pages branch if missing
+        run: |
+          if ! git ls-remote --exit-code origin gh-pages; then
+            git checkout --orphan gh-pages
+            git rm -rf .
+            {
+              echo "theme: jekyll-theme-tactile"
+              echo "title: ashim"
+              echo "description: Helm chart repository for ashim — self-hosted ML image processing"
+            } > _config.yml
+            {
+              echo "## ashim Helm Chart"
+              echo ""
+              echo "Self-hosted image processing with background removal, inpainting, upscaling,"
+              echo "face enhancement, colorization, and OCR."
+              echo ""
+              echo "### Install"
+              echo ""
+              echo '```bash'
+              echo "helm repo add ashim https://ashim-hq.github.io/ashim"
+              echo "helm repo update"
+              echo "helm install ashim ashim/ashim --namespace ashim --create-namespace"
+              echo '```'
+              echo ""
+              echo "### GPU variants"
+              echo ""
+              echo '```bash'
+              echo "# NVIDIA"
+              echo "helm install ashim ashim/ashim --set image.tag=latest-cuda"
+              echo "# AMD"
+              echo "helm install ashim ashim/ashim --set image.tag=latest-rocm"
+              echo '```'
+              echo ""
+              echo "### Source"
+              echo ""
+              echo "[github.com/ashim-hq/ashim](https://github.com/ashim-hq/ashim)"
+            } > index.md
+            git add _config.yml index.md
+            git commit -m "chore: init gh-pages for Helm chart repo"
+            git push origin gh-pages
+            git checkout main
+          fi
+
+      - name: Run chart-releaser
+        uses: helm/chart-releaser-action@v1.6.0
+        env:
+          CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+
+      - name: Ensure Tactile theme and landing page on gh-pages
+        run: |
+          git fetch origin gh-pages
+          git checkout gh-pages
+          {
+            echo "theme: jekyll-theme-tactile"
+            echo "title: ashim"
+            echo "description: Helm chart repository for ashim — self-hosted ML image processing"
+          } > _config.yml
+          git add _config.yml
+          git diff --cached --quiet || git commit -m "chore: refresh gh-pages theme config"
+          git push origin gh-pages || true
+          git checkout main
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -113,5 +113,5 @@ jobs:
           push: false
           tags: ashim:ci
           build-args: SKIP_MODEL_DOWNLOADS=true
-          cache-from: type=registry,ref=ghcr.io/${{ github.repository }}:cache-linux-amd64
+          cache-from: type=registry,ref=ghcr.io/${{ github.repository }}:cache-linux-amd64-cpu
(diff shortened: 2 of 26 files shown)


snapotter-hq:
Hey @swheettaos @stephenwheet , sorry this sat without a reply for so long - that's on me.

The root cause breakdown on the 29 GB image is spot on, and splitting into `latest` / `latest-cuda` / `latest-rocm` is the right design. Helm chart is a nice bonus too.

Unfortunately `main` has moved a lot since you opened this (17 conflicting files). The big changes on our side:

- Project renamed from ashim to SnapOtter (user, paths, branding, everything)
- AI models now install on demand per feature instead of bulk downloading via `download_models.py`
- Entrypoint picked up PUID/PGID support, venv bootstrapping, startup banner
- CUDA base bumped to 12.9.2 with digest pinning

That said, the stuff I want to keep from this PR: the `GPU_BACKEND` build arg for cpu/cuda/rocm pip splits, `ubuntu:24.04` as the base instead of nvidia/cuda (runtime injection is the way to go), three-tag manifests in the release workflow, `GITHUB_TOKEN` over `GHCR_TOKEN`, and the Helm chart.

Would you want to rebase onto current `main`? I can help walk through the conflicts if that's useful. Or if you'd rather not deal with all 17 files, I can take these ideas and wire them up myself - full credit to you in the commits and release notes either way.

Thanks for putting the time into this.

PR #224, opened 2026-06-12
feat: 2.0 phase 5 wave 4 - office, ebooks, data, archives (14 tools)

snapotter-hq:
## Summary

14 new tools (catalog 127 to 141) across four families:

- **Office (LibreOffice):** excel-to-pdf, powerpoint-to-pdf, and format converters for documents (docx/odt/rtf/txt), spreadsheets (xlsx/ods/csv), and presentations (pptx/odp). The doc engine gained explicit filter-name targets, and the image gained libreoffice-impress (it previously shipped writer and calc only, discovered by the new pptx fixture's acceptance gate).
- **Markdown and ebooks (pandoc, new ~215 MB engine):** markdown-to-html, markdown-to-docx, epub-convert (docx/html/md, plus pdf through the existing WeasyPrint bridge), and to-epub. Calibre was evaluated and rejected: 5.2 GB installed with 328 dependencies including Qt WebEngine, plus a no-sandbox requirement; mobi/azw3 are deferred with that rationale.
- **Data (pure JS):** merge-csvs (header-set validation, delimiter preserved), yaml-json (safe schema only, both directions by extension), xml-to-csv (first repeating element list, rows count in the envelope). The three existing CSV tools now accept .tsv end to end.
- **Archives:** create-zip (multi-file) and extract-zip with full zip-bomb defenses: central-directory size/count/ratio checks before any byte is decompressed, traversal and symlink rejection, and stream-level size enforcement against lying archives.

Security posture:

- pandoc never fetches remote resources: resource embedding is disabled on every call site, and a crafted remote-ref epub is proven to pass URLs through as text (html) or fail with the standard rejection (pdf).
- The wave-2 SSRF pre-scan now also catches single-slash scheme URLs (the form pandoc emits when rewriting epub content), restoring the friendly rejection for dirty books; the data:-only fetcher remains the backstop.

## Test plan

- [x] Local gates: typecheck 9/9, lint 0 errors, full suite 11,036 passed / 192 skipped (LibreOffice/pandoc/python suites are Docker-gated by design; data and archive suites run everywhere)
- [x] In-container compose smoke: 19/19, including the Impress pptx proof, the epub-to-pdf chain, both SSRF proofs (passthrough on html, rejection on pdf), the zip traversal guard, and the TSV acceptance check
- [x] In-container engine verification task: full LibreOffice conversion matrix, pandoc version/flag detection, and the epub chain verified before any route was built

[...]

PR #221, opened 2026-06-12
feat: 2.0 phase 5 wave 3a - video depth (22 tools)

snapotter-hq:
## Summary

22 new video tools (catalog 91 to 113), all ffmpeg-backed and fully locally testable (both the local and container ffmpeg builds carry every needed filter, so no skip gates anywhere this wave):

- **Geometry:** resize-video (presets + custom), crop-video (probed-bounds validation), rotate-video, change-fps, video-color
- **Temporal:** video-speed (pitch-preserving atempo chain or classic pitch-shift using the probed sample rate), reverse-video (5-minute memory guard), video-loudnorm (EBU R128)
- **Aspect:** aspect-pad (letterbox to 16:9 / 9:16 / 1:1 / 4:3 / 3:4) and blur-pad (blurred-bars background fill)
- **Overlay:** watermark-video (drawtext via textfile and an explicit resolved font, expansion disabled so user text is always literal), stabilize-video (two-pass vidstab, long-running)
- **Conversion:** gif-to-video, video-to-webp (animated), video-to-frames (all / every-Nth / timestamps, zipped, with frame caps)
- **Multi-input:** merge-videos (normalizing concat with silence fill for soundless inputs), replace-audio (codec-aware output container), burn-subtitles, embed-subtitles (mp4/mkv container rule), images-to-video (slideshow)
- **Utility:** extract-subtitles, video-metadata (strips metadata and reports what was found in the result envelope)

Framework work:

- Per-position input kinds on the media handler (video, audio, image, subtitle) so mixed-input tools validate each file correctly; subtitle files validate by extension, size, and content shape without ffprobe
- Font resolver in the media engine (static ffmpeg builds have no fontconfig; the container uses the DejaVu fonts shipped since wave 2)
- Multi-input staging and shared ffmpeg progress helpers; probed sample rate added to media info

Frontend: 22 settings components, five new multi-file tools, display modes, and i18n verified across all 21 locales.

## Test plan

- [x] Local gates: typecheck 9/9, lint 0 errors, full suite 10,912 passed / 146 skipped (only the wave-2 Docker-gated suites skip locally; every wave-3a suite runs)
- [x] In-container compose smoke: 17/17 PASS, including the font-resolution proof (watermark rendered inside the container), the embed-subtitles mkv branch, the merge-videos silence-fill branch, the four-tool subtitle chain (burn, embed, extract against a real mkv), frame-zip and animated-webp magic bytes, and the metadata report in the sync envelope

[...]

PR #219, opened 2026-06-11
feat: 2.0 phase 4 - wave 1 tools across all modalities

snapotter-hq:
## Summary

- **17 new tools** (5 video, 3 audio, 5 PDF incl word-to-pdf, 4 data) growing the catalog from 53 to 70, all running on the phase-3 modality framework
- **Factory multi-input** (`maxInputs`) with single-request multi-file uploads from the web app; merge-pdf accepts N PDFs in one POST
- **doc-engine**: qpdf merge/split/rotate + ghostscript compress wrappers; profile-isolated LibreOffice for word-to-pdf
- **Still-image-as-video discriminator**: MediaInputHandler rejects still images uploaded to video tools via container/probe analysis
- **Per-tool integration suites** incl the first 202+poll long-hint pattern (word-to-pdf); hostile-input matrix extended to all new modalities
- **E2E specs** for media-player and document display modes
- **Tool counts now derive from the catalog** (single source of truth); locale sweep covers all 21 locales
- **acceptedInputs extension gate** for document/file modality tools: factory rejects files with wrong extensions before the input handler
- **stripInternalPaths on worker SSE error frames**: terminal failed frames no longer leak scratch/tmpdir paths
- **pdf-to-image reclassified** as document modality (it consumes PDFs, not images)
- In-container word-to-pdf proof: POST .docx -> 202 {jobId} -> poll -> 200 with %PDF- header

Stacked on #218; merge order: #216 -> #217 -> #218 -> this.

## Test plan

- [x] `pnpm typecheck` clean (9/9 packages)
- [x] `pnpm lint` clean (86 pre-existing warnings, zero errors)
- [x] `pnpm test` full suite: 291 files, 10,685 tests pass, 0 failures, 83 skips (engine-gated), 3 files skipped
- [x] `pnpm check:license-boundary` green
- [x] `pnpm --filter @snapotter/web build` green
- [x] Docker compose smoke (SKIP_MUST_CHANGE_PASSWORD=true):
  - [x] resize: 200, image processed
  - [x] mute-video (tiny.mp4): 200, 13290 -> 8243 bytes
  - [x] rotate-pdf (test-3page.pdf, angle:90): 200, preview generated
  - [x] csv-json (tiny.csv): 200, 33 -> 137 bytes
  - [x] word-to-pdf (tiny.docx): 202 -> poll -> 200, %PDF- header confirmed (soffice in container)
- [x] Branch: 21 conventional commits, working tree clean, no local-only files tracked

PR #218, opened 2026-06-11
feat: 2.0 phase 3 - modality framework (metadata, handlers, engines, display modes)

snapotter-hq:
## Summary

- **Shared modality metadata on Tool + five-modality registry** -- every tool declares its modality (image/media/document/ai/utility); registry maps modalities to pool config, caps, and handler.
- **InputHandler split** -- image chain extracted verbatim into its own handler; media handler adds capped ffprobe validation (duration/resolution/codec caps from env); document handler adds capped qpdf validation (page-count caps from env). 1112-case format-matrix proof across the full suite.
- **Ref-based v2 process contract with legacy adapter** -- workers receive `InputRef[]` and load all inputs from object storage; legacy tools transparently adapted. Registry stores resolved copies; dead piscina pool removed.
- **Pool routing by modality + executionHint** -- sync window semantics: `long` execution hint skips the sync fast-path entirely. Pool selected by tool modality.
- **@snapotter/media-engine scaffold** -- ffmpeg spawn wrapper, capped ffprobe, progress parsing, encoder seam. skipIf-gated tests (require ffmpeg binary).
- **@snapotter/doc-engine scaffold** -- qpdf wrappers and profile-isolated LibreOffice conversion. skipIf-gated tests (require qpdf/soffice).
- **Python dispatcher profiles** -- docs instance with pikepdf page count alongside existing AI instance.
- **Per-modality previews** -- video poster frame via ffmpeg, PDF first-page render via qpdf. PreviewGenerator split by modality.
- **FileEntry modality + previewKind** -- file entries carry their modality and preview kind for frontend routing.
- **media-player + document display modes** -- PDF.js-based document viewer with error recovery, page reset, render cancellation; media player shell. No tool drives them until Phase 4.
- **Catalog regrouped by modality** -- tool catalog sections organized by modality above categories.
- **Dockerfile ships engines** -- ffmpeg 8 static binaries + qpdf + LibreOffice + pikepdf/PyMuPDF, +710 MB accepted.
- **Per-modality env caps** -- MEDIA_MAX_DURATION_S, MEDIA_MAX_RESOLUTION, DOC_MAX_PAGES configurable from environment.

## Test plan

- [x] `pnpm typecheck` -- 9/9 packages pass
- [x] `pnpm lint` -- 8/8 tasks, zero errors (86 pre-existing warnings)
- [x] `pnpm test` -- 273 files passed, 2 skipped; 10604 tests passed, 65 skipped; zero failures (336.81s)
[...]

PR #225, opened 2026-06-12
feat: 2.0 phase 5 wave 5a - image gap-fill (11 tools)

snapotter-hq:
## Summary

11 new image tools (catalog 141 to 152), closing every genuine gap a registry diff found against the catalog's image section, plus 4 enhancements to existing tools. One new dependency (bwip-js); everything else is sharp and hand-rolled SVG.

- **Adjustments:** pixelate (block size, region-capable API), duotone (two-color affine mapping), vignette (radial gradient)
- **Geometry:** circle-crop (transparent PNG out, pixel-verified), image-pad (canvas to aspect with color)
- **Generators:** barcode-generate (six symbologies via bwip-js, mirrors the qr generator's no-upload route), chart-maker (bar/line/pie from csv/json rendered to hand-rolled SVG with mandatory XML escaping of user labels, no chart library)
- **Utilities:** lqip-placeholder (tiny blurred webp + data URI in the envelope), histogram (256-bin per-channel render + channel stats), sprite-sheet (grid composite with a frame-coordinates payload), gif-webp (animated conversion both directions)

Enhancements: convert now preserves animation for gif/webp targets and gains ppm/eps/tga outputs through the hardened ImageMagick shim; qr-generate embeds an optional logo (error correction forced to H under occlusion); favicon ships a true multi-size .ico container with a graceful fallback when ImageMagick is absent.

Fixes shipped in-branch: the route-drift exemption for the new no-upload generator, and the in-container ImageMagick policy now grants EPS write (read|write for EPS only; PS/PDF/XPS stay read-only).

## Test plan

- [x] Full integration suite: 6,798 passed / 202 skipped, zero failures (drift test green post-fix)
- [x] Unit suite fully green at 152 tools; typecheck 9/9; lint 0 errors
- [x] In-container compose smoke: 15/15 including pixel-level proofs (circle alpha corners, pad corner color), the EPS write re-proof after the policy fix, the multi-size ICO header check, and the qr logo byte-difference check
- [x] SVG-injection regression: a chart label containing script tags renders as escaped text and the output decodes as a valid PNG

Stacked on #224. Merge order: #216 -> #217 -> #218 -> #219 -> #220 -> #221 -> #222 -> #224 -> this.


snapotter-hq:
Squash-merged to main as part of the 2.0 train merge.

PR #266, opened 2026-06-17
fix: post-2.0 audit findings (404 route, worker logging, engine cleanup)

snapotter-hq:
## Summary

Small, self-contained fixes surfaced while auditing the codebase for the 2.0.0 release. Each was verified against source before changing.

| Area | Fix |
|------|-----|
| web | Add a catch-all `*` route rendering `NotFoundPage`. Unknown URLs previously rendered a blank screen (the page existed but was never routed). |
| api | Worker `error` handler now uses `logger.error({ err, pool })` instead of `console.error`, so non-system pool errors flow through structured logging. |
| ai | Gate the `outpaint` script via `object-eraser-colorize` in the Python `TOOL_BUNDLE_MAP` (it was the only ML script missing a bundle gate; matches `inpaint`). |
| doc-engine | Resolve the `pandoc` binary through `PANDOC_PATH`, matching the `qpdf` / `ghostscript` / `libreoffice` / `pdfcpu` wrappers. |
| test | Drop the stale `tests/unit/landing/**` vitest exclude (the directory no longer exists post-Astro migration). |
| landing | Remove three unreferenced components (`HowItWorks`, `ModalityExplorer`, `TerminalBlock`); confirmed zero imports and no glob usage. |

## Verification

- `pnpm typecheck` passes (9/9 workspaces)
- `biome check` clean on changed files
- `apps/landing` build succeeds (165 pages) after the component removals

## Notes

- Branched from `main`, so it does not include the in-progress `fix/ai-bundle-build-bugs` work.
- One reported item was a false positive and intentionally left unchanged: `enhance_faces` is NOT broken under the `upscale-enhance` bundle (that bundle ships the `mediapipe` pip package, the legacy `mp.solutions` API needs no model file, and the fallback auto-downloads the model).


snapotter-hq:
Superseded by two focused PRs so the bug fixes can land independently of the judgment-call cleanup:
- #267 - fix: bug fixes (404 route, worker logging, outpaint gate, pandoc path)
- #268 - chore: dead-code cleanup (3 unused landing components + stale vitest exclude)

PR #351, opened 2026-06-24
test(e2e): modernize the GUI e2e suites for the 2.0 UI

snapotter-hq:
## Why

The nightly E2E jobs (E2E Full shard 3, Cross-Browser, Device Matrix) have been red on pervasive selector/premise drift left over from the 2.0 UI redesign. This sweeps the GUI e2e suites to match the current app and fixes a couple of real product issues surfaced along the way.

## What

**gui-tools-* suites** (essential, format, expanded, overlay, layout, metadata): settings and the submit button now mount only after upload (so "disabled without file" asserts the control is absent pre-upload), the result panel's revert control is "Adjust settings" not "Undo", several format/preset counts grew, some labels moved, and image/PDF fixtures live under `tests/fixtures/image|document`. Tools that render results inline without a review step (color-palette, image-to-pdf, favicon) keep their settings mounted, so those tests verify continued access rather than a non-existent undo.

**gui-cross-browser** (firefox + webkit): resize submits before checking download, settings opens via the avatar menu, theme-toggle title is lowercase, pipeline steps counted by Remove buttons, qr-generate targets the URL input. WebKit surfaces unauthenticated 401s as console errors, so the login/privacy render checks ignore "Failed to load resource". Dropped the fullscreen-grid test (the `/fullscreen` page and its toggle were removed in 2.0).

**Device matrix** (mobile + tablet): the tablet core flow expands the mobile settings bottom sheet before filling the width input and works on both iPad (810px) and Galaxy Tab S9 (640px). `@visual` device specs route to the chromium device projects only (baselines aren't browser-suffixed) and are excluded from the nightly device-matrix run since linux baselines aren't committed (same policy as the desktop visual specs).

**Shared / product:**
- `openSettings` helper rewritten to branch on viewport width and open Settings from the mobile bottom nav or the top-nav avatar dropdown (the 2.0 redesign removed the sidebar it used to click). Added `data-testid="user-menu"` to the avatar button so it works for any logged-in user.
- Real UI fix: a tall portrait overflowed the crop canvas because react-image-crop forces the preview img to `max-height:inherit`; the viewport cap now lives on the ReactCrop element so the child img inherits a real bound.
- e2e infra: point `FEATURE_MANIFEST_PATH` at a non-existent file so the API runs in native mode and reports AI bundles as available, instead of replacing each AI tool's UI with an install prompt the e2e env can never satisfy. AI GUI specs assert on settings only; processing specs still skip when the sidecar is absent.

## Verification (local)

Green: gui-tools-essential, -format, -expanded (134), -overlay, -layout, -metadata; gui-cross-browser on firefox + webkit (48); full device matrix (54 functional + 8 chromium visual).

Note: this does not yet cover the E2E Serial bucket, which has its own separate drift (tracked as follow-up).


snapotter-hq:

PR #354, opened 2026-06-28
feat(analytics): instance-wide opt-out for anonymous product analytics

snapotter-hq:
## Summary

Re-introduces a runtime, instance-wide admin opt-out for anonymous product analytics. It ships ON by default with no consent or onboarding page, and an admin turns it off in one click under Settings > System > Privacy. One toggle disables both PostHog and Sentry, no rebuild required. The `SNAPOTTER_ANALYTICS=off` build arg stays as a compile-time hard-off.

Analytics never carries filenames, file contents, OCR output, image metadata, or document text:

- Per-event property allow-lists on both the API and the browser. The raw search query is dropped from the payload and the type.
- Sentry is gated at the SDK boundary (`beforeSend` / `tracesSampler` / `beforeSendTransaction`), so auto-captured uncaught errors and transactions stop too, not just our explicit `captureException` calls. Crash reports are reduced to the error class plus a basename-only stack; message, logentry, request, extra, contexts, breadcrumbs, and hostname are stripped.
- The gate fails closed (transient DB error keeps a known-off state off; cold boot stays silent until the first successful read).

"Anonymous" is now literally true: `person_profiles: "identified_only"`, `instance_id` is no longer registered as a person property, and the public config endpoint blanks keys, DSN, and instance id when disabled.

Other pieces: cross-replica opt-out via Redis pub/sub (reuses the job-cancel channel pattern), a non-production `ANALYTICS_BAKED_OVERRIDE` seam so the toggle is testable in CI (the committed bake is off), i18n across all 21 locales, demo mock updated, and corrected docs, landing, and OpenAPI copy.

Also fixes a pre-existing bug found during e2e: the System settings Save button returned 400 for admins because it re-POSTed read-only and redacted keys (`instance_id`, masked secrets). Without this fix the new toggle could not be saved through the UI.

## Test plan

- `pnpm typecheck`, `pnpm lint`, `pnpm build`: green
- Analytics unit: 11 passing (gate truth table + fail-closed, property allow-list, behavioral no-leak)
- Analytics integration: 8 passing, shuffle-verified order-independent (effective truth table, blank-when-off, admin 200, non-admin 403)
- Analytics e2e: 9 passing, including an admin opting out via the real Save button and the effective config flipping off
- [ ] CI full suite on push

## Follow-up (not in this PR)

`AdminSecuritySettings.handleSave` has the same read-only/redacted bulk-save bug the System section had. Same one-line filter fixes it. Left out to keep this PR scoped to analytics.


snapotter-hq:

PR #355, opened 2026-06-28
feat: prebuilt pipeline templates in Automate

snapotter-hq:
## Summary

Adds a Templates section to the Automate tab with 9 curated, ready-to-run pipeline templates. Pick one, it loads into the existing builder with sensible preset settings, then you tweak it, run it, or save your own copy. The catalog is static shipped content in `packages/shared`, so there are no backend, schema, or migration changes: loading and running reuse the existing `/pipeline/execute` path.

## Templates

- Image: Web-optimize, Email-friendly, Privacy scrub, Social square, Watermark batch
- Documents & Files: Shrink PDF for email
- Audio: Podcast polish
- AI (model bundle required): Cut out subject, Restore old photo

Every template references real, pipeline-runnable tools with settings validated against each tool's live Zod schema.

## How it works

- `packages/shared/src/pipeline-templates.ts`: `PIPELINE_TEMPLATES` plus a `templateRequiredBundles()` helper that derives required model bundles from a template's steps, so nothing is hand-maintained.
- Names and descriptions go through a new `pipelineTemplates` i18n namespace, translated across all 21 locales.
- `TemplateCard` and `TemplatesSection` render the cards grouped by modality above Saved Pipelines; "Use template" calls the builder's existing `loadSteps`.
- An integration drift guard boots the app and asserts every template's tool IDs are registry-runnable and every settings object passes the live schema, so a future tool rename or schema change fails CI instead of silently breaking a template.

## Fixes (surfaced by loading preset settings)

- `compress-settings`: loading a compress or compress-pdf step with non-default settings drove an infinite render loop. Switched to a one-time seed plus an idempotent emit guard. The standalone Compress and Compress-PDF tools are unchanged.
- `remove-bg-settings`: the control ignored its incoming settings, so the Cut out subject template silently dropped its transparent-WebP, edge-refine, and decontaminate preset. It now seeds local state from the incoming settings once on load. Standalone Remove Background is unchanged.

## Scope notes

- Templates render in the desktop Automate panel, matching where Saved Pipelines lives (the mobile panel carries only the tool palette).
- AI templates show a badge per required bundle; installing a bundle still uses the existing feature-not-installed prompt at run time.

## Test plan

- [ ] `pnpm typecheck`
- [ ] `pnpm vitest run tests/unit/pipeline-templates.test.ts tests/unit/web/template-i18n.test.ts tests/unit/web/i18n-parity.test.ts`
- [ ] `pnpm vitest run tests/integration/pipeline-templates-drift.test.ts`
- [ ] `pnpm playwright test tests/e2e/automate-templates.spec.ts`


snapotter-hq:

PR #502, opened 2026-07-11
feat(tools): remove background from animated GIFs (WebP, APNG)

### LINKED ISSUE #496: Remove background from animated GIFs
### Problem

A user searched in-app for a way to **remove the background from an animated GIF** and found no matching tool (surfaced through the search-miss feedback prompt). The existing remove-background tool targets still images, so animated GIFs aren't covered.

### Proposed solution

Support background removal on animated GIFs: split the GIF into frames, run background removal per frame, and reassemble, outputting a transparent animation in an appropriate format (transparent GIF, APNG, or WebP, whichever preserves alpha best).

Considerations:

- Per-frame consistency (avoid flicker in the matte between frames).
- Preserve frame timing/loop metadata.
- Performance: many frames on CPU could be slow (relates to the GPU-acceleration work).

### Alternatives considered

- Document a pipeline recipe (extract frames -> remove-background per frame -> reassemble) if that is achievable with today's tools, as an interim answer.
- Limit to short GIFs / frame caps initially.

### Additional context

Surfaced as a "no results" search, which is a signal of latent demand: users expect this and can't find it.

---
_Filed by a maintainer from community feedback._

snapotter-hq:
Closes #496.

Adds a dedicated `remove-gif-background` AI tool that removes the background from an animated image (GIF, animated WebP, APNG) frame by frame and reassembles a transparent (or composited) animation. The still `remove-background` tool is untouched.

## What it does

- Detects animation, runs one warm rembg session over every frame, applies the chosen effect per frame, and re-encodes while preserving frame timing and loop.
- Output formats: animated WebP (default, full alpha), APNG (full alpha), and GIF (1-bit alpha, with an in-UI caveat).
- Per-frame effects: transparent, solid color, gradient, blurred-original background, image replacement, and drop shadow.
- Reuses the existing `background-removal` bundle, so anyone who already has it installed can use this immediately.
- Closes the search miss that filed the issue via new keywords.

## Notable implementation details

- **Sharp can't see APNG frames** (libvips `pngload` reports no pages), so animation detection is format-routed: Sharp `metadata().pages` for GIF/WebP, a small `acTL`-chunk parser for PNG/APNG.
- **The whole frame loop runs in one Python script** (`gif_remove_bg.py`). PIL reads all three formats and writes animated alpha output, one rembg session avoids per-frame model reloads, and it sidesteps libvips' page-height limitation on assembled frames. No ffmpeg dependency.
- **Frame coalescing is disposal-aware** (a naive `seek()` garbles disposal-optimized GIFs and animated WebP).
- **Model, device, and matting are locked once per animation**; the birefnet to u2net OOM fallback happens at animation scope, never per frame, to avoid matte flicker.
- **The sidecar timeout is scaled by frame count** (the default would SIGTERM a long run and take the shared dispatcher down with it).
- **Cancel is plumbed via a sentinel file** the Python loop polls between frames, since `ctx.signal` isn't wired to Python today, so cancelling a long job stops it instead of wasting the whole animation.
- A configurable `GIF_BG_MAX_FRAMES` cap (default 150) is enforced at the route before enqueue.

## Verification

Full monorepo typecheck (9/9), a stubbed-rembg Python smoke test covering all three output formats plus every effect and cancel, the route integration test (install gate / not-animated / over-cap / background-image-required), and every drift and catalog-integrity guard pass. i18n parity across all 21 locales, the OpenAPI spec, and the tool-count updates (241 to 242) are green.

## Remaining acceptance step

CI can't run the ML models, so end-to-end matting on a real GIF is a live GPU-box step, the same as every prior AI tool. The rembg integration point is identical to the proven still tool. Known v1 limitations, documented in the settings UI and landing copy: temporal flicker on complex motion, GIF's hard 1-bit edges, and per-frame cost that scales with length.


snapotter-hq:
## End-to-end verification with the real model

Ran the full pipeline locally against the actual rembg `u2net` model (CPU, not stubbed) on a real 4-frame animation: a person composited onto a busy seascape, panning per frame so every frame is distinct.

Result: the background is genuinely removed on every frame, across all output formats.

- All three formats produced valid 4-frame output and returned `{"success": true, "frames": 4}`: WebP (RGBA), APNG (RGBA), GIF (palette).
- Alpha check on the transparent WebP: background corners alpha = `[0, 0, 0]` (removed), subject-center alpha = `254` (kept), 72.9% of pixels fully transparent.
- The color-background effect composited the subject over a solid color correctly.
- Per-frame timing (`duration=[200, 250, 200, 150]`) and the loop were preserved.

So the AI matting, per-frame loop, disposal-aware coalescing, and the WebP/APNG/GIF encoders are verified end-to-end. GPU is now only a performance consideration, not a correctness unknown.

PR #544, opened 2026-07-16
fix(ai-bundles): stop CPU onnxruntime from clobbering onnxruntime-gpu

### LINKED ISSUE #490: AI GPU acceleration silently falls back to CPU: onnxruntime (CPU) overwrites onnxruntime-gpu in the AI venv
### Before submitting

- [x] Filed by a maintainer from community feedback
- [x] Searched existing issues (related but distinct from the iGPU/AMD requests #190, #126, #171, #124, #120)

### What happened?

On a working NVIDIA + CUDA host (reported running under Unraid with `--gpus all` and the NVIDIA Container Toolkit installed), AI tools run on **CPU instead of GPU**. The sidecar logs the fallback:

```
[python] GPU detected by torch but CUDAExecutionProvider not available in onnxruntime -- install onnxruntime-gpu for GPU acceleration
[python] No GPU detected, processing on CPU
```

This matches the code path in `packages/ai/python/gpu.py`: `torch.cuda.is_available()` succeeds, but `_try_onnx_cuda()` finds that `onnxruntime.get_available_providers()` does not contain `CUDAExecutionProvider`, so ONNX-based tools drop to CPU.

### Root cause (diagnosed by the reporter, and consistent with our code)

Two conflicting ONNX Runtime packages end up installed in the same AI venv (`/data/ai/venv`):

- `onnxruntime` (CPU-only), pulled in transitively as a dependency of **faster-whisper** (the transcription bundle)
- `onnxruntime-gpu` (e.g. `1.20.1`), installed for GPU acceleration

Both packages install into the **same directory**, `site-packages/onnxruntime/`. When the CPU `onnxruntime` is installed *after* `onnxruntime-gpu` (which happens because faster-whisper's dependency resolution adds `onnxruntime` during AI bundle install), it **overwrites the GPU build's files**. The pip metadata for `onnxruntime-gpu` survives, so `pip show onnxruntime-gpu` still lists it, but the module actually imported at runtime is the CPU build. Consequently `get_available_providers()` resolves to something like `['AzureExecutionProvider', 'CPUExecutionProvider']` and never `CUDAExecutionProvider`.

In short: installing the transcription bundle alongside GPU tools silently downgrades ONNX Runtime to CPU-only.

### Workaround (confirmed working by the reporter)

Reinstall the GPU build last, inside the running container:

```bash
docker exec -it SnapOtter /data/ai/venv/bin/pip uninstall onnxruntime onnxruntime-gpu -y
docker exec -it SnapOtter /data/ai/venv/bin/pip install onnxruntime-gpu==1.20.1
```

After this the GPU is used correctly.

### Expected behavior

When a CUDA-capable GPU is present and the transcription bundle (faster-whisper) is installed alongside ONNX-based GPU tools, `onnxruntime` should resolve to the GPU build and `CUDAExecutionProvider` should be available. Installing one bundle should not silently break another's GPU acceleration.

### Suggested directions

- Reconcile ONNX Runtime after installing bundles so a transitively-pulled CPU `onnxruntime` cannot clobber `onnxruntime-gpu`. Options: never let both coexist, install/repair `onnxruntime-gpu` **last**, or pin/constrain the transcription bundle so it doesn't drag in the CPU `onnxruntime`.
- Add a post-install verification: when a GPU is present, assert `onnxruntime.get_available_providers()` contains `CUDAExecutionProvider`, and self-heal (reinstall `onnxruntime-gpu`) or emit a loud, actionable warning if not.
- Extend the existing cross-bundle install safeguards to cover this ONNX CPU-vs-GPU conflict, the same way ABI conflicts are already guarded.

### Affected area

AI sidecar / on-demand bundle install pipeline (`packages/ai`, `install_feature.py`, `/data/ai/venv`). Impacts any ONNX-backed GPU tool (for example `remove-background` via rembg) once the transcription bundle is also installed.

### How are you running SnapOtter?

Docker on Unraid, NVIDIA GPU, `--gpus all`, NVIDIA Container Toolkit installed.

### Version

Not specified by reporter (reported early July 2026).

### Additional context

The reporter was very positive about the project and offered this diagnosis to help improve the Docker image.

---
_Filed by a maintainer from community feedback. Reporter contact details intentionally omitted._

snapotter-hq:
## Summary

Fixes #490.

Both PyPI onnxruntime flavors unpack into the same `site-packages/onnxruntime/` directory. Bundles are venv-delta tarballs merged file-by-file into the shared `/data/ai/venv`, so installing the transcription bundle (which carries a CPU `onnxruntime` via faster-whisper's transitive dependency) after any GPU bundle overwrote the GPU build's native libraries. The stale `onnxruntime_gpu` dist-info survived, so `pip show` kept claiming the GPU build was installed while `CUDAExecutionProvider` was gone and every ONNX-backed tool silently ran on CPU.

## The fix: GPU wins, in both install orders

`install_feature.py` gains `reconcile_onnxruntime()`, called inside the venv-writing breadcrumb window right before the merge:

- An incoming CPU onnxruntime is dropped from the staged bundle when a GPU build exists, so it can never clobber CUDA. The GPU build ships `CPUExecutionProvider` too, so it satisfies every consumer of the CPU build; the reverse is not true.
- An incoming GPU build clears the venv's stale CPU metadata. This also means **reinstalling any GPU bundle (e.g. Background Removal) repairs a venv that was clobbered before this guard existed**, using only the published bundle host, no runtime PyPI access.
- `gpu.py`'s fallback warning now tells users exactly that, instead of suggesting a manual `pip install`.

Build-side (takes effect on the next manual bundle publish):

- `build-bundle.sh` reconciles the flavor after pip installs, so a bundle can never ship CPU files under GPU metadata (rembg and faster-whisper both drag in the CPU flavor next to a pinned `onnxruntime-gpu`).
- `verify-bundle-compatibility.sh` now layers bundles through the real installer merge (reconcile + `move_tree`) instead of a raw `cp -a`, and asserts the merged venv ends up with exactly one flavor, with the CUDA provider library present when that flavor is GPU. Same spirit as the numpy-ABI cross-bundle check this script was born from.

## Testing

- New pytest suite `packages/ai/python/tests/test_install_feature_onnx_reconcile.py` (6 cases: both install orders, both no-op orders, empty venv, both-flavors-in-one-bundle). Full Python suite: 182 passed.
- Two end-to-end cases in `tests/unit/features/install-feature-prebuilt.test.ts` running the real installer against crafted bundle tars: clobber prevention and the repair path. Written first, watched fail, then wired the call site.
- `bash -n` on both shell scripts; the compat-script merge snippet smoke-tested standalone.
- Live GPU verification against the actual published bundles on an NVIDIA box is running next; results will be posted as a PR comment.

### DIFF
--- a/docker/build-bundle.sh
+++ b/docker/build-bundle.sh
@@ -252,6 +252,38 @@ for patch in patches:
 print("  Source patches applied")
 PYPATCH
 
+# ── Step 4.6: ONNX Runtime flavor hygiene ────────────────────────────────
+# A pinned onnxruntime-gpu plus a transitive CPU onnxruntime (rembg and
+# faster-whisper both depend on plain `onnxruntime`) leaves BOTH dist-infos in
+# site-packages, with the package files belonging to whichever installed last.
+# Force the GPU build's files back and drop the CPU metadata, so a bundle can
+# never ship CPU files under GPU metadata, or two flavors at once (#490).
+echo "=== Reconciling ONNX Runtime flavor ==="
+python3 << 'PYONNX'
+import os, shlex, shutil, site, subprocess, sys
+
+site_packages = site.getsitepackages()[0]
+names = os.listdir(site_packages)
+cpu = sorted(n for n in names if n.startswith("onnxruntime-") and n.endswith(".dist-info"))
+gpu = sorted(n for n in names if n.startswith("onnxruntime_gpu-") and n.endswith(".dist-info"))
+
+if not (cpu and gpu):
+    print("  Single ONNX Runtime flavor (or none); nothing to reconcile")
+    sys.exit(0)
+
+version = gpu[-1][len("onnxruntime_gpu-"):-len(".dist-info")]
+cmd = (f"{sys.executable} -m pip install --no-cache-dir --force-reinstall "
+       f"--no-deps onnxruntime-gpu=={version}")
+print(f"  > {cmd}", flush=True)
+if subprocess.run(shlex.split(cmd)).returncode != 0:
+    print("ERROR: onnxruntime-gpu reinstall failed", file=sys.stderr)
+    sys.exit(1)
+
+for name in cpu:
+    shutil.rmtree(os.path.join(site_packages, name), ignore_errors=True)
+print(f"  Kept onnxruntime-gpu=={version}; removed CPU dist-info: {', '.join(cpu)}")
+PYONNX
+
 # ── Step 5: Download models ──────────────────────────────────────────────
 echo "=== Downloading models ==="
 python3 << 'PYMODELS'
--- a/docker/verify-bundle-compatibility.sh
+++ b/docker/verify-bundle-compatibility.sh
@@ -77,7 +77,28 @@ for bundle_id in ${BUNDLE_IDS}; do
   mkdir -p "${staging}"
   tar -xzf "${archive}" -C "${staging}"
   if [[ -d "${staging}/site-packages" ]]; then
-    cp -a "${staging}/site-packages/." "${SITE_PACKAGES}/"
+    # Layer through the REAL installer merge (reconcile + move_tree), not a raw
+    # cp: install_feature.py reconciles the ONNX Runtime flavor between bundles
+    # (#490), and this script must verify the state users actually end up with.
+    # INSTALL_FEATURE_PY can point at a repo checkout mounted over an older
+    # image (same pattern as the install_runtime.py mounts in ai-bundles.yml).
+    "${VENV}/bin/python3" - "${staging}/site-packages" "${SITE_PACKAGES}" <<'PYMERGE'
+import importlib.util, os, sys
+
+script = os.environ.get("INSTALL_FEATURE_PY", "/app/packages/ai/python/install_feature.py")
+spec = importlib.util.spec_from_file_location("installer_under_verify", script)
+mod = importlib.util.module_from_spec(spec)
+spec.loader.exec_module(mod)
+if not hasattr(mod, "reconcile_onnxruntime"):
+    print(f"FAIL: {script} has no reconcile_onnxruntime; mount the current "
+          "packages/ai/python/install_feature.py into the container "
+          "(-v $PWD/packages/ai/python/install_feature.py:/app/packages/ai/python/install_feature.py:ro)",
+          file=sys.stderr)
+    sys.exit(1)
+staging_sp, site_packages = sys.argv[1], sys.argv[2]
+mod.reconcile_onnxruntime(staging_sp, site_packages)
+mod.move_tree(staging_sp, site_packages)
+PYMERGE
   fi
   rm -rf "${staging}"
   echo "  Installed ${bundle_id}"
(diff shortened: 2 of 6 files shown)


snapotter-hq:
## Live GPU verification (RTX 4070, real published bundles)

Ran the full matrix on an NVIDIA box against `snapotter/snapotter:latest` and the actual `deepsafe/feature-bundles` artifacts, driving `install_feature.py` directly inside the container.

**1. Baseline (stock installer):** install `background-removal` into a fresh `/data/ai/venv`:

```
providers: ['TensorrtExecutionProvider', 'CUDAExecutionProvider', 'CPUExecutionProvider']
site-packages: onnxruntime/  onnxruntime_gpu-1.20.1.dist-info/
```

**2. Bug reproduction (stock installer):** install `transcription` on top:

```
providers: ['AzureExecutionProvider', 'CPUExecutionProvider']
site-packages: onnxruntime/  onnxruntime-1.27.0.dist-info/  onnxruntime_gpu-1.20.1.dist-info/
```

Exactly the reporter's symptom, and worse than expected: faster-whisper's unpinned dependency has drifted to CPU onnxruntime **1.27.0**, so the clobber is also a cross-version mix. The stale `onnxruntime_gpu` dist-info keeps claiming the GPU build is installed.

**3. Repair path (patched installer):** reinstall `background-removal` on the clobbered venv:

```
[install] replacing the CPU build of onnxruntime with the GPU build; removed its stale metadata
providers: ['TensorrtExecutionProvider', 'CUDAExecutionProvider', 'CPUExecutionProvider']
site-packages: onnxruntime/  onnxruntime_gpu-1.20.1.dist-info/
```

This is the documented recovery for anyone already hit: upgrade, then reinstall any GPU bundle.

**4. Clobber prevention (patched installer):** reinstall `transcription`:

```
[install] kept the GPU build of onnxruntime: dropped this bundle's CPU onnxruntime so it cannot disable CUDA for other AI tools (#490)
providers: ['TensorrtExecutionProvider', 'CUDAExecutionProvider', 'CPUExecutionProvider']
transcription-imports-ok; ct2-cuda-devices: 1
site-packages: onnxruntime/  onnxruntime_gpu-1.20.1.dist-info/
```

CUDA survives, metadata stays truthful, and transcription itself still works end to end (faster-whisper and ctranslate2 import, CT2 sees the GPU).

snapotter-hq:
Follow-up: the build-side scripts from this PR are now also **execution-verified on real artifacts** (they'd previously only been syntax-checked, since they run inside the bundle-build container):

**`build-bundle.sh` Step 4.6** ran inside a full real build of `background-removal` for amd64-gpu (4.8 GB tarball produced, exit 0). It took the no-op path there, because current rembg no longer drags in CPU onnxruntime. The reconcile branch itself was then exercised for real against `/opt/venv` seeded into the worst-case state (CPU installed after GPU, both dist-infos present): it force-reinstalled `onnxruntime-gpu==1.20.1`, removed the CPU dist-info, and the venv ended importing the GPU build with `CUDAExecutionProvider` available.

**`verify-bundle-compatibility.sh`** ran end to end over the freshly built background-removal bundle plus the published `transcription-amd64-gpu.tar.gz` (checksum-matched against the manifest). The real-installer merge fired the reconcile while layering transcription, the numpy/scipy/scikit-image constraint checks still pass, and the new flavor check reported:

```
=== Checking ONNX Runtime flavor consistency ===
  ONNX Runtime flavor: GPU (onnxruntime_gpu-1.20.1.dist-info), CUDA provider library present
PASS: All bundles for amd64-gpu are mutually compatible
```

Also closed the remaining runtime gap: a full **remove-background inference through the HTTP API** on the RTX 4070 with both bundles installed (the previously-broken state) completed with the dispatcher logging `[gpu] CUDA available via ONNX Runtime + nvidia-smi: NVIDIA GeForce RTX 4070` and 42% GPU utilization / 6.5 GB VRAM during the run. Nothing was published anywhere; the locally built tarball was deleted after verification.

One test-harness note for future sessions: installing bundles via `docker exec` (as root) leaves model files unreadable by the app user, which shows up as a scrubbed `Permission denied: '[internal]'` job error, and it bypasses the post-install `shutdownDispatcher()`, so the old dispatcher keeps its cached no-GPU probe. Both are artifacts of the bypass, not product bugs; `chown -R snapotter:snapotter /data/ai` plus a container restart reproduces the real post-install state.

PR #613, opened 2026-07-21
fix(download): guard against stream/Content-Length size mismatch

### LINKED ISSUE #590: [Bug] image-to-PDF (png-to-pdf) download starts but never finishes on some self-hosted deployments
## Summary

A user reported (v2.0.0) that image-to-PDF conversion (`png-to-pdf`) hangs on download: "the download creates a file, you hit download, the download starts, but never finishes." It happened for every conversion regardless of file count or type, and clearing cache/cookies did not help. The server logic and download endpoint are heavily tested, but every test buffers the body via `app.inject`, so the streaming download over a real socket is never exercised. The signature points at the streamed `GET /api/v1/download/...` desyncing at the connection layer (declared `Content-Length` vs bytes actually delivered), most plausibly through a reverse proxy or a storage backend whose `stat()` size disagrees with the readable stream.

## Where it lives

- Preset: `packages/shared/src/conversion-presets.ts:194` (`png-to-pdf` -> base `image-to-pdf`).
- Base tool metadata: `packages/shared/src/constants.ts:149-160` (`executionHint: "fast"`).
- Route: `apps/api/src/routes/tools/image-to-pdf.ts`. This is a **custom inline route** (`registerImageToPdfRoute`, line 106), not `createToolRoute`. It does all work in the HTTP handler, does not use BullMQ, and does not insert a `jobs` row. `buildPdf()` (256-289) uses pdfkit; `putObject("outputs/<jobId>/images.pdf", ...)` (302); returns JSON `{ downloadUrl }` (304-311).
- Download endpoint: `apps/api/src/routes/files.ts:87-157`. Sets `Content-Length` by hand from `getObjectSize()` (107, 154) then `reply.send(await getObjectStream(key))` (155). Range branch 124-146.
- Object storage: `apps/api/src/lib/object-storage.ts` (`getObjectStream` 171-181 returns `createReadStream`; `getObjectSize` 283-290 returns `stat().size`).
- Global header mutation: `apps/api/src/index.ts:387-396` (`onSend` sets CSP/security headers on every response, including the streamed download).
- Frontend: `apps/web/src/components/tools/image-to-pdf-settings.tsx` submits via its own XHR (239, 300s timeout), then downloads with a plain `<a href={downloadUrl} download>` anchor (395-405), matching the shared `triggerDownload` other tools use.

## Root causes (ranked)

1. **Content-Length vs delivered-bytes desync on the streamed GET (most likely for a self-hosted deployment).** The handler sets `Content-Length` from `stat()` and streams an `fs.ReadStream`, and the global `onSend` mutates headers after the handler. If a reverse proxy (nginx, Traefik, Cloudflare) re-frames the response, or the declared length ever exceeds bytes delivered, the browser stalls at "started but not finished." This is the classic signature: 100% reproducible per environment, independent of file type or count.
2. **`stat().size` disagrees with the readable stream on some storage volumes.** If the deployment's volume (network or overlay FS) reports a `stat()` size larger than the stream yields, the GET waits forever for a tail that never arrives. Backend-specific; lower probability than 1.
3. **Front-end anchor/CSP edge (least likely).** The anchor mirrors the working `triggerDownload`; if other tools' downloads also fail in the same environment this collapses into cause 1.

## Reproduction / next datum

The fix branches on the deployment topology, so gather that first:

1. Deploy v2.0.0 as the reporter runs it (Compose vs all-in-one, and any reverse proxy). Upload one PNG, create the PDF, click Download.
2. In DevTools Network, compare the `GET /api/v1/download/<jobId>/images.pdf` response `Content-Length` against bytes received, and whether it ends or hangs in "receiving." Received < Content-Length confirms cause 1/2.
3. Isolate proxy vs app: `curl -v -o out.pdf "<origin>/api/v1/download/<jobId>/images.pdf"` through the proxy, then again directly against the API container port. Hangs through the proxy but completes direct means cause 1.
4. Capture `STORAGE_MODE` (local vs s3) and the `WORKSPACE_PATH` filesystem type.

## Suggested fix

- [ ] Stop setting `Content-Length` by hand for the local backend. Either hand Fastify / `@fastify/static` `sendFile` and let it own length + range semantics, or ensure the declared length always equals bytes sent. The current split (manual `Content-Length` + global `onSend` header mutation + raw stream body) is the seam where a proxy or backend can desync framing.
- [ ] Add an integration or e2e test that consumes the download over a **real socket** (not `app.inject`) and asserts received byte count equals `Content-Length`, for both the collated PDF and the ZIP path. This is the coverage gap that let it ship.
- [ ] Document that reverse proxies must not buffer and must pass `Content-Length` through unchanged for `/api/v1/download/`.

## Test gap

`tests/integration/tools/image/image-to-pdf.test.ts` and `tests/integration/platform/download-range.test.ts` are thorough but GET the download only via `app.inject`, which fully buffers the body, so a real-socket stream/Content-Length hang cannot surface.

harshjainnn:
## What does this PR do?
Adds a byte-count integrity check and inactivity timeout to the streamed 
download handler in `files.ts`, so a stat()/stream size disagreement 
(the still-unconfirmed "cause 2" from #590) fails loudly with a logged 
error instead of hanging the connection indefinitely. Complements #604, 
which addressed reverse-proxy buffering (cause 1) but left this storage-
backend case with no detection or observability.

Fixes #590

## Type of change
- [ ] New feature (non-breaking change that adds functionality)
- [ ] New translation or i18n update
- [ ] Documentation update
- [x] Test improvement
- [ ] Refactor (no functional change)

## Checklist
- [x] I have signed the CLA (the bot will prompt you if not)

## Screenshots (if applicable)
N/A — backend-only change, no UI impact.

## Notes for reviewers
This does not change happy-path behavior, headers, or the existing Range 
request branch — it only wraps the non-range full-stream path with:
1. A running byte count compared against the declared `Content-Length` 
   on stream end, logged as an error on mismatch.
2. A 30s inactivity timeout on the stream that destroys the response 
   socket and logs a timeout error, rather than hanging silently.

New integration test mocks `getObjectSize()` returning a size larger than 
what the mock stream actually emits, and asserts the request aborts 
within the timeout window instead of hanging.

Follow-up not covered here: `X-Accel-Buffering: no` (#604) only helps 
nginx-family proxies. If the #590 reporter turns out to be behind 
Traefik/Caddy/Cloudflare, we'll likely need chunked transfer encoding 
for the local storage backend as a separate PR.

### DIFF
--- a/apps/api/src/routes/files.ts
+++ b/apps/api/src/routes/files.ts
@@ -157,7 +157,59 @@ export async function fileRoutes(app: FastifyInstance): Promise<void> {
           `attachment; filename="${encodeURIComponent(filename)}"; filename*=UTF-8''${encodeURIComponent(filename)}`,
         )
         .header("Content-Length", String(size));
-      return reply.send(await getObjectStream(key));
+
+      const stream = await getObjectStream(key);
+
+      const INACTIVITY_TIMEOUT_MS = 30000;
+      let bytesStreamed = 0;
+      let timeoutId: NodeJS.Timeout | null = null;
+
+      const cleanup = () => {
+        if (timeoutId) {
+          clearTimeout(timeoutId);
+          timeoutId = null;
+        }
+      };
+
+      const resetTimeout = () => {
+        cleanup();
+        timeoutId = setTimeout(() => {
+          cleanup();
+          request.log.error(
+            { key, declaredSize: size, bytesStreamed },
+            "Stream inactivity timeout: no data emitted within timeout window",
+          );
+          reply.raw.destroy();
+        }, INACTIVITY_TIMEOUT_MS);
+      };
+
+      resetTimeout();
+
+      stream.on("data", (chunk: Buffer | string) => {
+        bytesStreamed += typeof chunk === "string" ? Buffer.byteLength(chunk) : chunk.length;
+        resetTimeout();
+      });
+
+      stream.on("end", () => {
+        cleanup();
+        if (bytesStreamed !== size) {
+          request.log.error(
+            { key, declaredSize: size, bytesStreamed },
+            "Stream size mismatch: emitted bytes do not match declared Content-Length",
+          );
+          reply.raw.destroy();
+        }
+      });
+
+      stream.on("error", () => {
+        cleanup();
+      });
+
+      reply.raw.on("close", () => {
+        cleanup();
+      });
+
+      return reply.send(stream);
     },
   );
 
--- a/tests/integration/platform/download-range.test.ts
+++ b/tests/integration/platform/download-range.test.ts
@@ -1,4 +1,5 @@
-import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import { afterAll, beforeAll, describe, expect, it, vi } from "vitest";
+import * as objectStorage from "../../../apps/api/src/lib/object-storage.js";
 import { deletePrefix, putObject } from "../../../apps/api/src/lib/object-storage.js";
 import { buildTestApp, type TestApp } from "../test-server.js";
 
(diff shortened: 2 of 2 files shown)


harshjainnn:

harshjainnn:

harshjainnn:
recheck

snapotter-hq:
Thanks for the report and the PoC, @harshjainnn. Your read of cause 2 was right, and it's the useful half of this.

I reproduced it over a real socket: when `getObjectSize` over-reports, the client sits on keep-alive framing waiting for a tail that never arrives, and hangs. That happens at the app layer, no proxy required.

The runtime guard needed rework, though. Counting with `stream.on("data")` before `reply.send` flips the source into flowing mode, and this repo's async `onSend` hook defers Fastify's pipe, so bytes can drain past the response and truncate the download. The 30s inactivity timer also resets on data events only, so a slow or backgrounded client trips it and a healthy transfer gets cut mid-stream.

Same instinct, reworked: #617 resets the socket on a shortfall through a backpressure-safe counting `Transform`, and adds the real-socket test at the generic download route that the buffered `app.inject` tests couldn't reach. You're credited via `Co-authored-by`.

Closing this in favor of #617. Thanks again for chasing it down.

PR #857, opened 2026-08-18
fix(api): stream tool outputs over 1.5 GiB instead of crashing on the 2 GiB read cap

### LINKED ISSUE #841: fix(api): tool outputs over 2 GiB fail with ERR_FS_FILE_TOO_LARGE; the worker buffers whole results
Tool outputs over 2 GiB kill the job with `RangeError: ERR_FS_FILE_TOO_LARGE`. Sentry NODE-2Z: 7 events since 2026-07-19 at `apps/api/src/jobs/worker.ts:375`, the `readFile(result.scratchPath)` call.

**What happens today.** When a tool's process function returns a `scratchPath` instead of a buffer, the worker reads the whole file into one Buffer before uploading it to object storage (`worker.ts:372`-398, same pattern for `extraOutputs` at line 391). Node's `fs.readFile` hard-caps a single read at 2 GiB, so any output past that throws. Video conversions are the realistic path here: a long 4K transcode or images-to-video render clears 2 GiB without trying. The user waits through the whole (long) encode and then gets a failure at the copy step, after the actual work succeeded.

Even under the cap it's a problem: a 1.9 GiB result means a 1.9 GiB heap spike per concurrent job.

**Fix sketch.** `putObjectStream()` already exists (`apps/api/src/lib/object-storage.ts:192`) and `assertLocalCapacity` is already wired into it. For scratchPath results, stream:

```ts
const size = (await stat(result.scratchPath)).size;
await putObjectStream(outputKey, createReadStream(result.scratchPath));
```

and keep the buffer path only for tools that return in-memory buffers.

**Gotchas, from reading the surrounding code:**

- `processedSize` accounting currently comes from `resultBuffer.length`; take it from `stat()` instead.
- Preview generation downstream consumes the result buffer for some modalities; for the streaming path, feed it the scratchPath (previews never need the full video in memory anyway).
- `autoSaveToLibrary` (issue #495 machinery) also touches the result; check whether it re-reads the buffer or the stored object.
- S3 backend: check what `putGenericObject` does with streams; may need the multipart upload path for >5 GiB.

**How found / why out of scope.** Sentry triage session 2026-08-18, review-only pass. The fix touches the worker result contract and needs a big-file integration test, too large for a triage session.

snapotter-hq:
Sentry NODE-2Z: `RangeError: ERR_FS_FILE_TOO_LARGE` at the worker's `readFile(result.scratchPath)`. A 4K transcode clears Node's 2 GiB single-read cap without trying, so the job did all the encoding work and then failed at the copy step.

The decision lives in a new `resolveOutputSource` (`apps/api/src/jobs/output-resolve.ts`): a scratchPath result under 1.5 GiB buffers exactly as before, so previews, the PDF producer scrub, and library auto-save are untouched for every size that used to work. Over the cap, the file stays on disk and streams to object storage through the existing `putObjectStream` (which already runs the capacity guards and returns the byte count, used for `processedSize`/`bytesOut`). Applies to the primary output and `extraOutputs` both.

Streamed outputs ship without a poster preview and skip library auto-save: both paths need the bytes in memory, and these jobs previously could not complete at all. The auto-save skip logs with jobId and size. Multi-GiB intermediates inside pipelines are unchanged by this PR (separate buffering path).

Tests: five unit cases on `resolveOutputSource` (buffer passthrough, under-cap buffering, over-cap streaming with a lowered cap injected, the neither-source error, and a pin that the default cap stays under 2^31). `tests/unit/api/jobs/` all pass (156 tests), `tsc --noEmit` clean, and the resize + trim-video integration specs run green through the real worker (23 tests, exercising both the buffer and scratchPath flows end-to-end).

Fixes #841

### DIFF
--- a/apps/api/src/jobs/output-resolve.ts
+++ b/apps/api/src/jobs/output-resolve.ts
@@ -20,7 +20,7 @@ export type ResolvedOutput =
 
 export async function resolveOutputSource(
   out: { buffer?: Buffer; scratchPath?: string },
-  label: string,
+  missingSourceMessage: string,
   opts: { maxBufferedBytes?: number } = {},
 ): Promise<ResolvedOutput> {
   if (out.buffer) {
@@ -33,5 +33,7 @@ export async function resolveOutputSource(
     }
     return { kind: "buffer", buffer: await readFile(out.scratchPath), size };
   }
-  throw new Error(`${label} returned neither buffer nor scratchPath`);
+  // The caller supplies its full historical message; integration tests and
+  // logs pin these strings per call site.
+  throw new Error(missingSourceMessage);
 }
--- a/apps/api/src/jobs/output-resolve.ts
+++ b/apps/api/src/jobs/output-resolve.ts
@@ -1,0 +1,37 @@
+/**
+ * Decide how a tool's output travels to object storage.
+ *
+ * Node's fs.readFile refuses single reads at 2 GiB (ERR_FS_FILE_TOO_LARGE),
+ * so buffering every scratchPath result crashed long video jobs after the
+ * encode itself had succeeded (Sentry NODE-2Z). Anything over the cap stays
+ * on disk and is streamed with putObjectStream; everything under it buffers
+ * exactly as before, keeping previews, PDF scrubbing, and library auto-save
+ * untouched for the sizes they actually serve.
+ */
+
+import { readFile, stat } from "node:fs/promises";
+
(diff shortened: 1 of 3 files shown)


snapotter-hq:
The Integration (4/4) failure was the worker-branches suite pinning the exact missing-output message for extras ('has neither', which the new resolver had reworded to 'returned neither'). Pushed a follow-up: resolveOutputSource takes the caller's full historical message, and the worker-branches spec passes locally again (15/15).

Issue #923, opened 2026-08-27
i18n: the editor menu bar renders English over a fully translated editor.menu.* tree


snapotter-hq:
## What happens today

The image editor's menu bar renders in English no matter what locale you pick. Open `/editor` with the UI set to German and File, Edit, Image, Layer, Select, Filter and View all read as English, along with every item under them.

The translations already exist. `editor.menu.*` in `packages/shared/src/i18n/en.ts` carries the whole tree (7 menus, the submenus, the adjustment and filter entries) and all 20 other locale files carry the translated version. It has been there since the namespace was authored. `apps/web/src/components/editor/editor-menu-bar.tsx` just never used it: `useMenuDefinitions` builds its `MenuDef[]` out of 87 hardcoded `label: "..."` strings.

A repo-wide search for `editor.menu.` finds three consumers, none of them the menu bar:

- `components/editor/editor-right-panel.tsx:128` -> `editor.menu.view.panels`
- `components/editor/options/transform-options.tsx:144,153` -> `editor.menu.edit.transform.flipHorizontal` / `flipVertical`

So roughly 85 translated strings x 21 locales are sitting unused while the UI beside them renders English. This is the same wiring bug as #647 (the help dialog), one order of magnitude larger.

## How it was found

Sweeping `components/editor` for #909. The scanner cannot see these labels because they live in object literals read back through a property access, which is the general blind spot tracked separately. The menu bar gets its own issue because it needs no new keys and no translation batch, so it is by far the cheapest large i18n win left in the app.

## Why it was out of scope

#909's manifest was the set of scanner hits. The menu bar produced exactly one hit, the `title="Back to SnapOtter"` attribute, which was fixed. Restructuring the whole `MenuDef` builder was not in that manifest and would have meant editing the largest editor file in a PR that already touched 77.

## Repro

1. `pnpm dev`, open `http://localhost:1351/editor`.
2. Switch the language to anything but English (footer language picker).
3. The menu bar is still File / Edit / Image / Layer / Select / Filter / View.

Expected: the German build shows Datei / Bearbeiten / Bild / Ebene / Auswahl / Filter / Ansicht, which are already in `packages/shared/src/i18n/de.ts`.

## Starting sketch

`useMenuDefinitions` is already a hook and the file already imports `useTranslation`, so this is mostly mechanical.

[...]

Issue #922, opened 2026-08-27
i18n: the drift guard cannot see labels held in data structures


snapotter-hq:
## What happens today

The drift guard from #906/#909 only sees strings that sit in a rendered position in JSX, or that reach one through a local variable. It is blind to the far more common pattern in this codebase: a label held in a data structure and read back through a property access.

```tsx
const SAMPLE_SIZES: { label: string; value: SampleSize }[] = [
  { label: "Point (1x1)", value: 1 },
  { label: "3x3 Average", value: 3 },
  { label: "5x5 Average", value: 5 },
];
// ...
{SAMPLE_SIZES.find((s) => s.value === sampleSize)?.label}
```

That renders English in all 21 locales and `pnpm vitest run tests/unit/web/tool-ui-literal-drift.test.ts` stays green. `tests/helpers/tool-ui-literals.ts` documents the limit in its header comment, but documenting it does not stop the drift.

A crude probe (count `label`/`title`/`description`/`hint`/`tooltip`/`placeholder`/`text` string literals inside object literals in `apps/web/src/**/*.tsx`) returns 422 across the app: 205 in `components/tools`, 111 in `components/editor`, 66 in `editor/panels`, 24 in `editor/common`, 9 in `components/settings`, 3 each in `components/common` and `editor/options`. That count is an upper bound, not a work estimate. Plenty of those never render, and some already resolve through a label map. But it is the right order of magnitude for the sweep.

There is a second, narrower blind spot in the same family: a literal handed straight to a setter or a non-JSX sink. `change-password-page.tsx:104` had `setError(data.error || "Failed to change password")`, which reaches the visible error paragraph, and `settings-dialog.tsx:2891` builds a `confirm()` message in English. Neither is an assignment nor a JSX child, so neither is reported.

## Confirmed instances, found while sweeping #909

These were read in context by the agents doing the sweep and left alone because they were outside the manifest. Each one renders English today.

- `components/editor/options/eyedropper-options.tsx:11-15` `SAMPLE_SIZES` labels, rendered at `:51`.
- `components/editor/tools/crop-tool.tsx` `ASPECT_RATIOS`, including `"Free"`. Harder than it looks: `"Free"` doubles as the lookup key in three places, so it needs an id split from the display string first.
[...]

Issue #919, opened 2026-08-24
test(e2e): two chromium specs fail on strict-mode violations, and `pnpm test:e2e` is red on main


snapotter-hq:
`pnpm test:e2e` does not pass on a clean `origin/main` checkout. Two chromium specs fail deterministically on Playwright strict-mode violations, and the Firefox and WebKit lanes fail on navigation errors. Neither is caught by PR CI, which runs only three chromium smoke specs.

Verified on `origin/main` at `03e0ea5a` in a fresh worktree, macOS, so none of this comes from a feature branch.

## The two deterministic ones

**`tests/e2e/gui-tools-ai.spec.ts:545`**, Smart Crop, "shows subject focus controls by default".

```
strict mode violation: getByRole('button', { name: 'Attention' }) resolved to 2 elements:
  1) <button aria-expanded="false" aria-label="Attention finds the most visually salient region. Entrop...
  2) <button class="flex-1 text-xs py-1.5 rounded ...">Attention</button>
```

The tooltip trigger's `aria-label` starts with the word "Attention", so it matches the accessible-name query for the strategy button sitting next to it. Both elements are real and correct; the selector cannot tell them apart.

**`tests/e2e/pdf-to-image-preset-multifile.spec.ts:62`**, "pdf-to-png reports the error".

```
strict mode violation: getByText(/All files failed processing/i) resolved to 2 elements:
  1) <p class="text-xs text-destructive-ink">All files failed processing</p>      (inside form)
  2) <p class="font-medium text-foreground mb-1">All files failed processing</p>  (Preview area)
```

Worth noting the app is behaving correctly here. The message the test wants does appear. It appears in two places, the form and the preview panel, and the assertion cannot pick one.

Both are selector problems rather than product bugs. Scope each to its container (`page.locator("form").getByText(...)`) or use `.first()` where either match is acceptable.

## The cross-browser lanes

Firefox and WebKit fail on navigation, not assertions: `NS_BINDING_ABORTED`, `NS_ERROR_FAILURE`, `Frame load interrupted`, and `Navigation to "..." is interrupted by another navigation`. A sample run of just `navigation.spec.ts` and `home-page.spec.ts` on `origin/main` gave 13 failed, 40 passed.

[...]

Issue #906, opened 2026-08-23
i18n: 16 more tool settings panels render bare English literals


snapotter-hq:
## What happens today

At least 16 tool settings panels render bare English literals for labels that should be locale strings. Switch the app to any non-English locale and these labels stay English:

- beautify-settings.tsx
- circle-crop-settings.tsx
- compress-settings.tsx
- conversion-preset-settings.tsx
- convert-settings.tsx
- erase-object-settings.tsx
- favicon-settings.tsx
- gif-tools-settings.tsx (line 211 renders a literal `Mode` while `toolSettings["gif-tools"].mode` exists in all 21 locales)
- image-to-base64-settings.tsx
- optimize-for-web-settings.tsx
- qr-generate-settings.tsx
- rounded-crop-settings.tsx
- split-settings.tsx
- stitch-settings.tsx
- svg-to-raster-settings.tsx
- watermark-text-settings.tsx

This list is a lower bound. It comes from grepping `apps/web/src/components/tools/*.tsx` for lines that are exactly a bare capitalized label (`^\s+(Mode|Width|Height|Quality|Format|Size|Color|Speed)$`). Longer literals (hints, sentences, multi-word labels) don't match that pattern, so the real count per file is higher. The gif-tools case shows both failure modes exist: some literals shadow keys that are already translated, others need new keys.

## How it was found

The sweep step of #904 (PR #905), which fixed the same defect in pixelate-settings, ImageViewer, MultiImageViewer, and the automate page. Sixteen more files was a campaign, not a rider on that diff.

## Expected

Every user-facing string goes through `useTranslation()`. Literals that shadow existing keys switch to those keys (no locale edits needed, like #810). Literals without keys get new ones in `packages/shared/src/i18n/en.ts` plus all 20 other locale files, with `pnpm i18n:check` verifying parity.

## Prior art worth cribbing

- PR #905: the whole recipe. Anchor-insert new keys after a unique neighboring key per locale file, crib existing per-locale wording where a term is already translated elsewhere (Mode/Width/Height live in the gif-tools and chart-maker blocks), then `biome check --write` the i18n dir. Its test (`tests/unit/web/i18n-hardcoded-strings.test.tsx`) shows the render-under-`I18nProvider`-with-stored-`de`-locale pattern, including the in-memory localStorage stub the jsdom env needs.
- #647 (help dialog) is the earlier instance of literals shadowing existing translations.

## Gotchas

[...]

PR #925, opened 2026-08-27
fix(web): route every rendered string in the web app through i18n

### LINKED ISSUE #909: i18n: extend the tool-UI literal sweep and drift guard beyond components/tools
## What happens today

The #906 sweep (PR #907) and its drift guard cover `apps/web/src/components/tools` only. Running the same scanner (`tests/helpers/tool-ui-literals.ts`) over the neighboring directories finds rendered English literals that show up in every locale:

- `apps/web/src/components/common`: 29 hits
- `apps/web/src/pages`: 65 hits
- `apps/web/src/components/editor`: 1 hit

Run `scanToolUiLiterals(<dir>)` to reproduce the lists with file/line/kind.

The scanner also has a known blind spot in any directory: a literal assigned to a variable before rendering (`const label = cond ? "Uploading..." : "Processing..."; ... {label}`) is invisible to its render-position analysis. Collage-preview shipped exactly this and was caught by luck during PR #908, not by the guard. More instances may exist undetected in the already-swept directory.

## How it was found

Closing out #906: the scanner was pointed at the other web directories while checking whether anything else remained. The blind spot came from PR #908's collage work.

## Why out of scope

#906 was scoped to tool settings panels; extending to three more directories plus a scanner capability is its own pass with its own translation batch.

## Prior art worth cribbing

PR #907 is the complete recipe: manifest generation, existing-key resolution (many literals will already have byte-identical keys), parallel wiring, per-locale translation, anchored locale insertion, and the drift-guard/allowlist pattern. PR #908 adds the data-parity test pattern. The two hard-won learnings from those PRs: grep can never prove an i18n key unreferenced (destructured locals hide access; the TranslationKeys typecheck is the only safe oracle), and `pnpm i18n:check`'s exit code is unreliable on main (compare counts against main instead, #882).

## Starting sketch

1. Extend `scanToolUiLiterals` to accept multiple directories, and teach it simple local-variable tracking: a `const x = <string-literal-bearing expression>` whose identifier is later rendered as a JSX child or user-facing attribute counts as a hit.
2. Re-run over tools + common + pages + editor; triage per the #907 manifest actions.
3. Wire, translate, assemble with the #907 machinery; extend the drift test's directory list so the guard holds everywhere it has swept.

snapotter-hq:
Fixes #909.

The #906 sweep stopped at `apps/web/src/components/tools`, and so did its drift guard. Everything outside that one directory kept shipping English to all 21 locales. This finishes the job: the scanner now sees more, the sweep covers all of `apps/web/src`, and the guard scans it recursively so a directory added next month is covered without anyone remembering to add it to a list.

## Scanner

Two capability changes, both TDD'd in `tests/unit/web/tool-ui-literal-scanner.test.ts` (14 cases against throwaway fixture files).

The blind spot the issue called out is closed: a literal assigned to a local before it renders (`const label = busy ? "Uploading..." : "Processing..."`) is now reported, resolved through nearest-scope lookup so an inner `const` shadows an outer one of the same name. It only follows render-transparent hops, so `items.filter(i => i.endsWith(".png"))` inside a rendered local stays quiet. Pointing that at the already-swept tools directory immediately turned up 6 hits PR #907 could not see, including the concatenated `Tiles will be very small (400x300px)` in split-settings.

`scanToolUiLiterals` also takes an array of directories now, recurses into subdirectories, and reports repo-relative paths instead of bare filenames so the failure output is clickable.

## The sweep

417 hits across ten directories: tools 6, common 30, editor 1, editor/common 59, editor/options 106, editor/panels 76, files 9, layout 11, settings 29, pages 68.

They resolved into 251 new keys, 68 distinct existing keys wired up, and 25 allowlist entries.

The reuse number is the interesting one. `automate.*`, `toolPage.*`, `settings.apiKeys.*`, `settings.auditLog.*` and `editor.menuBar.backTitle` were already authored and translated into 21 locales, sitting unreferenced while the JSX next to them rendered hardcoded English. 41 of the 43 hits in automate-page and tool-page were pure wiring. Same story for 26 of the 29 in settings-dialog.

Sentence fragments split around `{expr}` were collapsed into whole-sentence keys rendered through `format()` and `plural()`, never keyed as fragments. `Page {page} of {pages} ({total} entries)` is one key, not four. The privacy policy's two paragraphs that wrap an inline `<code>` stay one key each: a small `SentenceWithCode` helper splits the template on `{code}` and renders the element between the halves, so a translator can move the token where their grammar wants it.

English output is byte-identical. Every `en` value equals the literal it replaced, composed sentences render the same characters, and the two agents that touched prose verified it by re-composing the pre-change JSX through the new `format()` path rather than by eye.

## Translations

251 keys x 20 locales, each translated against its own existing file as the style reference so established terminology carries over instead of being re-picked. A validator checked path parity, placeholder parity per key, and dash rules on every batch before insertion.

## Guard

`tests/unit/web/tool-ui-literal-drift.test.ts` now scans `apps/web/src` recursively instead of one hardcoded directory. The companion assertion still fails on allowlist entries that no longer occur, so the list cannot rot.

The allowlist grew from 18 to 25 and its comment now says what belongs there: format and codec names, brand and license names, units, placeholder examples, and identifiers that only look like copy. That last category is the new one. `processed-file` is a download basename that `ReviewPanel` hands to `triggerDownload()`; translating it would localize the filename users get on disk. `admin` is the fallback username that also drives the avatar initial. `alt="background"` is a blurred backdrop layer stacked under a properly labelled subject image.

`tests/unit/web/i18n-web-sweep.test.tsx` adds locale-sentinel coverage for four surfaces from this sweep. Asserting on English proves nothing here, since the provider defaults to `en`, so every assertion is against the German bundle, plus a check that the raw `{placeholder}` never reaches the DOM. Mutation-checked: reverting one component to its literal turns the test red.

## What the pre-PR review changed

Two holes in the scanner, both in code this branch adds, and fixing the second one found a live bug.

Function parameters were invisible to the scope walk, so a prop named `label` shadowing an outer `const label = "..."` reported a string that never renders. That matters more than a stray false positive: the only way to silence a phantom offender is an allowlist entry, and the allowlist matches on text alone, so one would have exempted that exact string everywhere under `apps/web/src`.

Literals in the template argument of `format()` and `plural()` were dead ends too, since no call expression counted as render-transparent. This branch makes both helpers the standard way to compose a sentence carrying an expression, so the next hardcoded string here will very likely be written in exactly that shape. Pointing the fixed scanner at the tree immediately turned up `plural(tm.memberCount, "member", "members")` in settings-dialog, which the entire sweep had walked past. Now two keys.

Four smaller ones from the same review: `lqip-placeholder`'s `alt="LQIP preview"` is real screen-reader text rather than a locale-invariant token, so it left the allowlist for a key; `file-library-modal` had its own key made instead of borrowing `automate.importButton` from the pipeline-builder namespace; the three locale-name fallbacks read `find(code === "en")` rather than `[0]`, so reordering `SUPPORTED_LOCALES` cannot silently relabel the button; and `SentenceWithCode` now splits on the first `{code}` and keeps the remainder, so a locale that drops or repeats the placeholder loses no text. Three test assertions that could have passed vacuously were tightened.

## Two behaviour changes, called out because they are not byte-identical

`url-import-modal.tsx` called `format(t.urlImport.readyCount, { ready, total })` against a key that reads `"{count} ready"`. `format()` leaves unmatched placeholders alone, so the modal footer rendered the literal text `{count} ready` in every locale including English. The call site now passes `count`.

`login-page.tsx`, `footer.tsx` and `top-nav.tsx` each had `current?.nativeName ?? "English"`. A `nativeName` is never localized (the same dropdown shows "Español" and "日本語" whatever the UI language), so the fallback is now `supportedLocales[0]`, whose `nativeName` is "English". Rendered characters unchanged; the literal is gone from all three.

## Deliberate calls a reviewer might question

Three "Opacity" labels in the editor options bars reuse `editor.shapes.opacity` rather than getting their own key. Same namespace, same single word, and no translator would want them to differ.

`appLayout.footer.toggleTheme` is "Toggle Theme" and `appLayout.topNav.toggleTheme` is "Toggle theme". The casing genuinely differs in the shipped app and byte-identity won over tidiness.

`settings.tools.disabledCount`, `settings.roles.userCountBadge` and the three `toolPage.download*` keys are now unreferenced. The first two read `{count} tool(s)` / `{count} user(s)` while the screen says "3 tools disabled" and "3 users", so they could not be reused without changing English. Left in place rather than pruned across 21 files.

Rendered filename fallbacks are allowlisted, never keyed: `processed-file` reaches `triggerDownload()`, and the others feed an extension badge derived with `.split(".").pop()`. Where such a fallback is provably unreachable it became `""` instead, which is what `convert-settings` already did.

## Verification

- Full monorepo typecheck green, which is the real parity oracle: all 21 locales conform to `TranslationKeys`.
- `pnpm test:unit`: 494 files, 8625 passed, 1 skipped.
- `pnpm build` green.
- `pnpm lint` clean (the 565 warnings are pre-existing and untouched).
- `pnpm i18n:check` identical to `origin/main` at `e88a30f3`: the same 280 docs and 40 landing-ui stale units (#882). It does not read `packages/shared/src/i18n` at all.
- Playwright `chromium` and `chromium-serial`, the full local run: 2131 tests, 2023 passed, 103 skipped, 5 failed.

The five, honestly:

`gui-tools-ai` Smart Crop and `pdf-to-image-preset-multifile` are the two deterministic strict-mode violations #919 documents as red on a clean `origin/main`.

`gui-navigation` "selecting a language updates the button label" was mine. Localizing the language button's tooltip meant `button[title='Language']` stopped matching the moment the test switched to Spanish, so it failed on "element(s) not found" rather than on its assertion. The button now carries a `data-testid` and the test uses it. The other title-based locators in that file and in `i18n.spec.ts` stay in English throughout, so they still match.

`pixelate-display` selection mode and `gui-settings-people` role dropdown are both a 30-second wait inside a 30-second test timeout, which cannot survive parallel load. Neither is plausibly caused by this branch: it touches neither pixelate nor the API, and the `en.ts` diff is additions only, so no English label moved. Three other tests in the same people spec use the identical `getByRole("button", { name: /people/i })` and passed.

A focused re-run against the current tree of those two plus the settings, automate, file-library, theme, pipeline and i18n specs (which is also where the second and third commits land) is green: 285 passed, including all 21 locale-loading cases.

## Found while in here, not fixed

The scanner still cannot see labels held in data structures rather than JSX. `editor-menu-bar.tsx` renders 87 hardcoded menu labels while `editor.menu.*` sits fully translated and unreferenced, and several panels carry English in module-scope option arrays. Filed as #922 (the scanner gap, with the inventory) and #923 (the menu bar, which needs no new keys at all).


snapotter-hq:
Rebased onto ba7fc514 (five main merges had landed since the branch point, then three more mid-review) and re-verified the whole thing. One small fix on top, 2de8266d.

What I checked before pushing:

- Reproduced #909 on main first: the scanner reports 141 raw hits outside components/tools (29 common, 65 pages, 1 editor, plus layout, settings and files), and the old guard stays green through all of it.
- Guard teeth: with the branch's guard in place I checked apps/web/src back out at pre-sweep main. The drift test fails with the expected offender list, and passes again on the swept tree.
- Locale parity: all 21 bundles are key- and placeholder-identical to en (script over the flattened catalogs, 4041 keys), and the monorepo typecheck, which is the real parity oracle, is green. Spot-checked de, ja, ar, ru and zh-CN values for the new keys: real translations, placeholders intact.
- Reused-key byte identity spot checks (toolPage.filesCount, automate.*, files.*, settings.auditLog.paginationInfo, the reviewPanel download keys): each matches the literal it replaced.
- pnpm lint and pnpm typecheck green. pnpm i18n:check identical to main: the same 280 docs and 40 landing-ui stale units from #882, and it never reads packages/shared/src/i18n.
- Full pnpm test on the rebased branch: 862 files passed, 4 skipped, zero failures. The fresh clean-main baseline's only two failures are local scratch tests that aren't in the repo at all.
- Adversarial review (code-reviewer plus silent-failure-hunter) found one real hole, fixed in 2de8266d: the session-fetch catch in settings-dialog stored the English sentinel "unknown", which is truthy, so the new t.settings.general.roleUnknown could never fire on the one durable path that renders it. The catch stores an empty string now.

Dismissed from the review, with reasons: the missing placeholder-parity guard is #934, the scanner's data-structure blind spot is #922 (I added a note there about destructured locals), and the over-broad skip classes (ALL-CAPS with spaces, leading-ellipsis, text-global allowlist words) are now #977. A SCIM race red that surfaced mid-verification reproduced on clean main at d6f071c1 and is tracked as #980; it's gone on ba7fc514 for both main and this branch.

Issue #967, opened 2026-09-03
fix: user-limit SSO denials leave no audit trail


snapotter-hq:
Found while reviewing the #928 fix. When SSO auto-create is denied because MAX_USERS is reached, nothing is written to the audit log and the warn line carries no identity.

## Where

`apps/api/src/lib/external-auth-resolver.ts`, the auto-create denial branch (the `user_limit_reached` return). Every sibling denial dual-writes: `user_disabled` audits `${PROVIDER}_LOGIN_FAILED` at all three of its call sites, and `user_not_authorized` audits plus logs `{ externalId, email }`. The limit denial only does a bare `logger.warn("oidc auto-create blocked: user limit reached")`.

The register route has the same gap in a milder form: its 403 USER_LIMIT_REACHED response is not audited either. Worth deciding both at once.

## Expected vs actual

Org sits at MAX_USERS. Five new employees try to SSO in over a week; all bounce.

Actual: the audit log (the thing a compliance review actually reads) shows nothing. The API log has five identical warn lines that cannot be told apart.
Expected: an `OIDC_LOGIN_FAILED` / `SAML_LOGIN_FAILED` audit row per denial with `reason: "user_limit_reached"` and the sanitized externalId, and a warn with `{ externalId, email }` like the `user_not_authorized` path already does.

## Why out of scope for #928

Pre-existing gap in the same block #928 rewrote; that PR stayed pinned to the race per the one-purpose rule.

## Sketch

In the denial branch:

```ts
await audit(`${providerUpper}_LOGIN_FAILED`, {
  reason: "user_limit_reached",
  externalId: sanitizeAuditInput(String(externalId)),
});
```

plus `logger.warn({ externalId, email }, ...)`. Mirror of the `user_not_authorized` terminal at the bottom of the resolver. Tests: assert `state.auditCalls` in `tests/unit/api/external-auth-resolver-mutation.test.ts` (the harness already records audit calls) and the audit row in `tests/integration/platform/oidc-callback-coverage.test.ts`, which already drives this exact denial end to end.

Issue #966, opened 2026-09-03
fix: SCIM provisioning ignores MAX_USERS entirely


snapotter-hq:
Found while reviewing the #928 fix. SCIM provisioning never checks MAX_USERS. Not a race: there is no check at all, not even the racy count the register route had before #928.

## Where

`apps/api/src/routes/enterprise/scim.ts`, the POST /Users handler (insert around line 393). It validates the duplicate userName and inserts into `schema.users` with no cap check on any path.

## Expected vs actual

With MAX_USERS=5 and five users existing, an IdP provisioning run creates a sixth via `POST /api/v1/scim/v2/Users`.

Actual: 201, user created, nothing logged, nothing audited. The People page shows 6/5.
Expected: the create is refused with a SCIM error envelope once the cap is reached, same invariant the register route and SSO auto-create now enforce.

There is a real question hiding here: maybe the cap deliberately should not bind IdP-driven provisioning (the IdP is the source of truth in enterprise setups). If that is the decision, it needs to be written down in the MAX_USERS docs instead of being an accident.

## Why out of scope for #928

#928 was pinned to the count-then-insert race in the two paths that already enforced the cap. SCIM enforcing it at all is a behavior decision, not a race fix.

## Prior art

#928 added `apps/api/src/lib/user-limit.ts` with `userLimitReached(tx)`: takes the shared pg advisory xact lock, counts inside the caller's transaction. If the cap should bind SCIM, wrap the SCIM insert in `db.transaction` and call the helper, exactly like the register route does. Denial should use the SCIM error envelope (`urn:ietf:params:scim:api:messages:2.0:Error`) with status 403, not the plain JSON the register route returns.

## Gotchas

- #927 is touching the same handler for the duplicate-create race. Land one before starting the other or the diffs will fight.
- The SCIM suite builds users through helpers in `tests/integration/platform/scim.test.ts`; a cap test must count existing rows first, same pattern as `tests/integration/platform/user-limit-race.test.ts`.

PR #972, opened 2026-09-03
fix(api): stop concurrent user creates from exceeding MAX_USERS

### LINKED ISSUE #928: fix: concurrent registers can exceed MAX_USERS by one (count-then-insert race)
Found while fixing #900. The register route's user-limit check has the same check-then-act shape as the duplicate-username race, but with no unique constraint to backstop it: two concurrent POST /api/auth/register calls with different usernames both count MAX_USERS - 1 rows, both pass the check, both insert. The cap ends up exceeded by one (N-1 for N concurrent requests) with no error anywhere.

## Where

`apps/api/src/plugins/auth.ts:944-953`: `db.select().from(schema.users)` then a length check, followed by scrypt hashing and the insert. The await on `hashPassword` between check and insert gives the same wide race window that made #900 reproduce deterministically in its test.

Same shape in the OIDC auto-create path: `apps/api/src/lib/external-auth-resolver.ts:184-188` counts against `env.MAX_USERS` before its insert.

## Expected vs actual

Actual: user count can pass MAX_USERS silently.
Expected: at most MAX_USERS users; the race loser gets the existing 403 USER_LIMIT_REACHED.

## Why out of scope for #900

That fix was pinned to the duplicate-username insert; this race has no constraint to conflict on, so it needs a different mechanism, and the #900 PR stayed single-purpose.

## Sketch

A pg advisory lock around count+insert, taken only when MAX_USERS > 0, is probably the cheapest honest fix (mind the advisory-locks-are-per-database gotcha from PR #918 if tests share a cluster). Alternatives: serializable transaction, or re-count after insert and delete-and-403 on overflow. The cap is off by default (0 = unlimited) and nobody has reported an overshoot, so decide how much rigor it actually needs before reaching for the heavy options.

snapotter-hq:
Closes #928

Two concurrent POST /api/auth/register calls with different usernames both counted MAX_USERS - 1 rows, both passed the check, both inserted. Cap exceeded by one, no error anywhere. The scrypt hash between check and insert keeps the window wide open; reproduced on main with two concurrent requests against a live API (MAX_USERS=2, one existing user: both 201, count landed at 3). The OIDC/SAML auto-create in external-auth-resolver had the same shape.

The fix is the sketch from the issue: a pg advisory lock around count+insert. New `apps/api/src/lib/user-limit.ts` exports `userLimitReached(tx)`, which takes `pg_advisory_xact_lock(7421003)` (extending the 7421001/7421002 key namespace from migrate.ts) and counts users inside the caller's transaction. Both paths wrap the locked count and the insert in one `db.transaction`, so the race loser waits on the lock, recounts after the winner commits, and gets the existing 403 / `user_limit_reached` denial. MAX_USERS=0 skips the lock entirely. scrypt runs before the transaction so the lock window is two fast statements.

The register route also stops capturing env.MAX_USERS at module load and reads it per request. Identical in production (env never changes at runtime); it's what makes the cap reachable from tests.

Tests: `tests/integration/platform/user-limit-race.test.ts` races register vs register, auto-create vs auto-create, and register vs auto-create (that one pins both paths onto the same lock). All three fail on main ([201, 201] / [created, created] / cap overshot) and pass here, stable across repeated runs. The resolver unit-test db mock learned the transaction contract; canned-row order moved because the count now runs last.

Verified locally: pnpm lint and pnpm typecheck clean; the race file plus every caller suite green (auth-edge-cases, oidc-auth, oidc-callback-coverage, oidc-mfa-callback, saml-auth, saml-cache, saml-license-gate, role-authority, scim: 219 integration tests, and the four auth/resolver unit files: 93 tests). Full local pnpm test runs kept losing their testcontainers to Docker VM restarts on a machine that was running a second test workload at the same time; the furthest run passed 11,416 tests with zero assertion failures before the cascade of connection errors, so the full-suite gate rides on CI here. No web, i18n, or docker surface in the diff.

Tradeoffs, documented rather than coded around:

- `pg_advisory_xact_lock` has no lock_timeout. A transaction wedged mid-commit would queue capped creates behind it; that takes an already-broken DB connection, and the locked window is a COUNT and an INSERT.
- The recount assumes read committed (Postgres default). An operator forcing `default_transaction_isolation = repeatable read` would reopen the race; nothing in our shipped stacks does.
- Two same-username registers racing at exactly the cap now resolve 201 + 403 instead of 201 + 409. Both were denied before, both are denied now, and the sequential duplicate case still 409s first.

Follow-ups filed while in here: #966 (SCIM provisioning ignores MAX_USERS), #967 (limit denials write no audit row).

### DIFF
--- a/apps/api/src/lib/external-auth-resolver.ts
+++ b/apps/api/src/lib/external-auth-resolver.ts
@@ -1,10 +1,10 @@
 import { randomUUID } from "node:crypto";
-import { and, eq, sql } from "drizzle-orm";
+import { and, eq } from "drizzle-orm";
 import type { FastifyBaseLogger } from "fastify";
-import { env } from "../config.js";
 import { db, schema } from "../db/index.js";
 import { isDisabledRole } from "../permissions.js";
 import { auditLog, sanitizeAuditInput } from "./audit.js";
+import { userLimitReached } from "./user-limit.js";
 
 // ── Types ─────────────────────────────────────────────────────────
 
@@ -179,15 +179,6 @@ export async function resolveExternalUser(params: ExternalAuthParams): Promise<E
       return { user: null, action: "denied", deniedReason: "user_disabled" };
     }
 
-    // Check user limit
-    if (env.MAX_USERS > 0) {
-      const [countResult] = await db.select({ count: sql<number>`COUNT(*)` }).from(schema.users);
-      if (countResult && countResult.count >= env.MAX_USERS) {
-        logger.warn(`${provider} auto-create blocked: user limit reached`);
-        return { user: null, action: "denied", deniedReason: "user_limit_reached" };
-      }
-    }
-
     // The username scan can't close the race: two concurrent logins both
     // pass it before either insert commits (issue #927), so the insert
     // carries a conflict guard and the loser recovers below.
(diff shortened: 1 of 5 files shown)


snapotter-hq:
Rebased onto d6f071c1 (#971), which rewrote this same auto-create block into a conflict-guarded retry loop. The merge keeps both behaviors: the locked count and the guarded insert share one transaction, and when the create is refused for either reason the resolver still checks whether this same external identity just won a concurrent login before denying. Concretely: at one slot below the cap, two tabs of the same new user now resolve created + matched instead of created + user_limit_reached.

Re-verified after the rebase: 352 tests across the 17 affected and caller files pass on the merged code, and the three race tests still fail when the resolver and register route are reverted to d6f071c1 alone.

Issue #990, opened 2026-09-03
test: SCIM race tests flake with a license-gate 403 when forks run in parallel under load


snapotter-hq:
The SCIM race tests in `tests/integration/platform/scim.test.ts` (the #971-era "concurrent duplicate group creates return one 201 and one SCIM 409" and the newer rename races) flake when the file runs beside another fork on a loaded box. One of the two concurrent requests comes back 403 with the license-gate envelope instead of reaching its insert or update, so the pg-race harness times out with `only 1/2 writes blocked on "teams"` and outcome `403,201`.

Seen twice on 2026-09-03: once in a locked full-suite run on the #970 branch (a tree with no #968 changes, so the flake predates the rename work) and repeatedly in two-file runs (`teams.test.ts` + `scim.test.ts`, VITEST_MAX_FORKS=2) on a machine with load average 10+. Never reproduced standalone: the same test passes in ~600ms with `-t` filtering, and passed in full-suite runs when the box was quiet.

What the instrumentation showed during the wedge (pg_stat_activity polled every 2s, filtered to the fork's own database):

- the helper transaction idle-in-transaction holding `LOCK TABLE "teams" IN EXCLUSIVE MODE`
- exactly one contender parked (`insert into "teams" ...`, wait_event Lock/relation)
- no third backend at all: the second contender never issued any SQL in that database

So the missing contender's 403 was produced entirely in Node before any query, which points at `requireScimFeature` -> `isEnterpriseFeatureEnabled("scim")` returning false despite `mockEnterpriseFeatures(["scim"])` being active for the suite. Two candidate mechanisms, not yet distinguished:

- `isEnterpriseFeatureEnabled` does `await import("@snapotter/enterprise")` per call, and its failure path degrades to false. A transient vite-node module-request failure under load would 403 exactly one request. This path logs "enterprise module failed to load" or "isFeatureEnabled threw", but through the lazily imported file-transport logger, so nothing appears on test stdout; nobody has checked the file log yet. That is the first thing to do.
- The mocked dynamic import itself resolving to the real `@snapotter/enterprise` (which returns false, silently, with no log) when two requests import it concurrently under vi.doMock + vi.resetModules generations. Would also produce a log-free 403.

[...]

Issue #970, opened 2026-09-03
fix: concurrent mixed-case team creates bypass the case-insensitive name check


snapotter-hq:
`POST /api/v1/teams` enforces case-insensitive name uniqueness only in its pre-check SELECT (`LOWER(name) = LOWER($1)`). The unique constraint on `teams.name` is exact-case, and so is the `onConflictDoNothing` guard #927 added. Concurrent creates of `Foo` and `foo` both pass the pre-check and both insert: two 201s and a case-twin pair the API otherwise treats as duplicates (renames onto either casing get 409'd, SCIM group lookups match one of them by exact name).

Sequential requests are fine; only the concurrent window is open. #927 flagged this and deliberately deferred it, and the #927 PR closed the exact-case race only. Filing so the deferral outlives the closed issue.

## Fix sketch

A functional unique index, `CREATE UNIQUE INDEX ... ON teams (LOWER(name))`, plus mapping its 23505 to the existing 409 body. Needs a dedupe pass for installs that already hold case-twins (from this race or from pre-pre-check history), which is the reason it did not ride along: migration plus dedupe policy, not a code-only guard. Drizzle expresses this via `uniqueIndex().on(sql\`lower(name)\`)` in `apps/api/src/db/schema.ts`.

## Gotchas

- The SCIM Groups create path inserts into the same table but its pre-check is exact-case, so the index would newly 409 SCIM group names differing only by case from an existing team. Probably correct, but it is a behavior change for SCIM IdPs to note in release notes.
- The teams PUT rename has the same case-insensitive pre-check; the index covers it once the rename 23505 is mapped (see the rename-races issue).

Issue #991, opened 2026-09-03
fix: concurrent role renames onto one name still 500, the rename-race fix missed roles PUT


snapotter-hq:
Found during the #968 review pass (its code-reviewer flagged it): the roles PUT rename has the exact bug #968 fixed in teams PUT and the four SCIM update paths, and it wasn't in #968's enumerated site list, so the fix PR left it alone.

`apps/api/src/routes/roles.ts`, PUT /api/v1/roles/:id: pre-checks for a name collision (around line 207, returns 409 `{error: "Role name already exists", code: "CONFLICT"}`), then runs an unguarded `tx.update(schema.roles).set(updates)` (around line 257) inside a `db.transaction`. `roles.name` is unique (`apps/api/src/db/schema.ts:185`).

Expected vs actual: two concurrent renames onto the same role name both pass the pre-check; the loser should get the pre-check's 409. Actual: the 23505 unique violation escapes as a 500 plus a Sentry event, the same shape as #927/#968.

Sketch: reuse `isUniqueViolation` from `apps/api/src/lib/pg-errors.ts` (landed with the #968 PR). One wrinkle vs the five #968 sites: this update runs inside a `db.transaction`, so catch around the transaction call (or around the update and rethrow a sentinel) so the rollback still happens before replying 409. Test recipe: `raceUpdates("roles", 2, ...)` from `tests/helpers/pg-race.ts` with two custom roles renamed onto one target, asserting [200, 409] and the pre-check body; the teams PUT race test in `tests/integration/platform/teams.test.ts` is the direct template, and custom-roles.test.ts already has the create-race sibling to sit next to.

Why out of scope for #968: the issue enumerated teams PUT and the SCIM user/group updates only, and the PR stayed on the enumerated sites.

PR #992, opened 2026-09-03
fix(api): return 409 instead of 500 when duplicate-name renames race

### LINKED ISSUE #968: fix: duplicate-name rename races return 500 in teams PUT and SCIM user/group updates
The #927 fix closed the check-then-insert races on the five creation paths. The rename flavor of the same bug is still live: each of these pre-checks for a name collision, 409s if found, then UPDATEs onto a unique column with no conflict handling. Two concurrent renames onto the same name both pass the check, the loser hits the unique constraint, Postgres raises 23505, and the global handler turns it into a 500.

## Where

- `apps/api/src/routes/teams.ts` PUT /api/v1/teams/:id: case-insensitive pre-check, then `db.update(schema.teams).set({ name })` with no guard. `teams.name` is unique (`apps/api/src/db/schema.ts:48`).
- `apps/api/src/routes/enterprise/scim.ts` PUT /Users/:id: pre-check ("userName already taken"), then an unguarded users UPDATE. Around line 565 as of the #927 PR.
- Same file, PATCH /Users/:id: `updates.username` set around line 702, same unguarded UPDATE.
- Same file, PUT /Groups/:id and PATCH /Groups/:id: displayName renames UPDATE `teams.name` (around lines 992 and 1043).

Realistic triggers are the same as #927: IdP provisioning retries for SCIM, two admins renaming at once for teams.

## Expected vs actual

Actual: the race loser gets a 500 and a Sentry event.
Expected: the 409 the pre-check already returns.

## Why it wasn't in the #927 PR

That PR stayed on the creation paths the issue enumerated. UPDATEs also need a different mechanism: there is no `onConflictDoNothing` for UPDATE, so this needs either a try/catch that maps 23505 to the pre-check's 409, or an `UPDATE ... WHERE NOT EXISTS` shape with a rowCount check. The 23505 catch is probably the simpler read; that also reopens #900's note that there is no shared pg-code-to-HTTP helper, and with four more sites it may finally be worth one.

## Prior art

The #927 PR (branch `fix/927-creation-race-500s`) has the whole recipe: `tests/helpers/pg-race.ts` makes these races deterministic (for UPDATEs, lock the table the same way; parked UPDATEs show up in pg_stat_activity as `update \"teams\"`, so the helper's ILIKE pattern needs to take the verb as a parameter or match both). The teams and SCIM race tests in `tests/integration/platform/` are the template.

## Gotchas

- The teams rename pre-check is case-insensitive but the constraint is exact-case, same mismatch as the create path (see the separate case-twin issue).
- SCIM PATCH paths build `updates` incrementally; the 23505 can come from the username field while the rest of the patch is valid. The 409 should abort the whole patch, matching what the pre-check does today.

snapotter-hq:
Closes #968

The #971 fix covered the five creation paths; the rename flavor was still live. Teams PUT, SCIM Users PUT, and SCIM Groups PUT pre-check for a name collision, 409 if found, then UPDATE onto a unique column with no conflict handling, so two concurrent renames onto the same name both pass the check and the loser surfaces the 23505 as a 500. The two PATCH paths are worse: SCIM Users PATCH and Groups PATCH have no pre-check at all, so even a sequential rename onto a taken name 500s today.

There's no onConflictDoNothing for UPDATE, so the five sites catch the unique violation and return their pre-check's exact 409 body instead. The helper is lib/pg-errors.ts isUniqueViolation(): it walks the drizzle cause chain for SQLSTATE 23505 (depth-capped against cyclic causes) and everything else rethrows untouched. The issue floated a shared pg-code-to-HTTP helper; a boolean predicate turned out to be the right size, because the five sites return three different response envelopes and the catch is four lines each.

Since the lower(name) index from #970 is on main, the same catches also turn mixed-case rename races on teams and groups into 409s; before this PR those were the one new 500 window #970 documented.

Test infra: tests/helpers/pg-race.ts grows raceUpdates() next to raceInserts() (shared lock-and-park machinery, parked UPDATEs instead of INSERTs), and the parking deadline moves from 10s to 25s: the 10s deadline flaked under fork-parallel load, and 25s stays under vitest's 30s testTimeout so the helper's own diagnostic (which names the unparked contender's outcome) fires instead of a generic test-timeout kill. a separate box-load flake in the SCIM race tests (a license-gate 403 eating one contender) is diagnosed and filed as #990, it predates this change.

Tests: five races through raceUpdates (teams PUT, SCIM Users PUT and PATCH, SCIM Groups PUT), each formerly [200, 500], now [200, 409] with the loser's body pinned to the pre-check envelope; two sequential PATCH-onto-taken-name tests (Users, Groups) that were straight 500s; and a unit spec for isUniqueViolation including the cyclic-cause guard. Reverting the fix with the tests kept fails all seven.

Verified: lint, typecheck, teams + scim suites (110 tests) on top of a6ecbc1f, custom-roles + external-auth-race (the other pg-race callers, confirming the helper refactor), pg-errors unit spec. The local locked full run died partway when the box hit a machine-wide disk-full event (test containers lost their storage; ~1,600 green marks and zero test failures before the cascade), so CI's unit plus 4-shard integration matrix is the full-suite evidence for this PR.

One decided tradeoff: a SCIM Users PATCH that both deactivates and renames in one body revokes sessions before the rename 409s. The pre-#968 behavior was a 500 with the same side effect, and SCIM PATCH here has never been transactional; noted rather than restructured.


snapotter-hq:
Rebased onto current main (`1e8ac965`); the previous base `a6ecbc1f` was 11 days and 5 commits behind. Clean rebase, no conflicts, diff unchanged. The commit body said the pg-race parking deadline moves to 30s where the code moves it to 25s, so that line now matches the code.

Re-verified locally on the rebased branch:

- Teeth check: with the three source files reverted to main's version and the tests kept, all six behavior tests fail with the reported shape. Teams PUT, SCIM Users PUT, SCIM Users PATCH, and SCIM Groups PUT races come back `[200, 500]`; the two sequential PATCH-onto-taken-name tests come back 500. Restoring the fix turns all six green.
- `pnpm vitest run` over teams, scim, custom-roles, external-auth-race, and the pg-errors unit spec: 129 tests, 128 passed. The one failure was the #990 license-gate flake (`only 1/2 writes blocked on "teams"`, outcome `403,200`) in the Groups PUT race, running five files across parallel forks. scim.test.ts standalone: 86/86 green.
- `pnpm lint` exit 0 (572 warnings, all pre-existing .astro noise), `pnpm typecheck` clean across all 9 workspaces.

Baseline for the diff: CI on main @ `1e8ac965` is green. Nightly on that same SHA is red in four e2e jobs (cross-browser, device matrix, full 2/4 and 4/4), pre-existing and untouched by an API-only change.

Follow-up #991 (the roles PUT rename, the one site this PR's own review pass turned up) is next, stacked on this.

PR #993, opened 2026-09-03
fix(scim): trim group displayName so whitespace twins are duplicates (#988)

### LINKED ISSUE #988: fix: SCIM group names skip the trim the teams API applies, so whitespace twins slip every uniqueness check
The teams create and rename routes trim names at the Zod boundary (`teamBodySchema` in `apps/api/src/routes/teams.ts`), but SCIM group provisioning inserts `displayName` raw (`apps/api/src/routes/enterprise/scim.ts`, POST /Groups and the PUT/PATCH renames). An IdP that sends `"Engineering "` creates a team that renders identically to `"Engineering"` but is a different row.

The lower(name) unique index from #970 doesn't catch these: `lower('foo ')` and `lower('foo')` are different strings, so whitespace twins still slip past every uniqueness layer.

Expected: SCIM display names get the same trim the API applies, so `"Engineering "` either matches the existing team (409 on create, merge on lookup) or lands trimmed.
Actual: a whitespace twin is created; membership then splits across two teams depending on which caller provisioned it.

Where: `apps/api/src/routes/enterprise/scim.ts`, the `displayName` reads in POST /Groups, PUT /Groups/:id, PATCH /Groups/:id (and the same question applies to `userName` on the Users paths, which the register route's `validateUsername` may already reject; check before touching).

Why out of scope for #970: that fix was the case-twin race (migration + index + guards). Trimming is input normalization at a different boundary, and touching SCIM input handling deserves its own look at what IdPs actually send.

Sketch: `displayName.trim()` at the top of the three group handlers, reject empty after trim with the existing scimError 400. Decide whether a dedupe pass for existing whitespace twins is worth it (probably not; they're rare and visible). Prior art: the Zod `.trim()` in teams.ts, and the #970 tests in `tests/integration/platform/scim.test.ts` show the create/409 assertion pattern to copy.

Found during the #970 review pass (code-reviewer flagged the asymmetry).

thejdubb02:
Fixes #988.

The teams create and rename routes trim names at the Zod boundary (teamBodySchema), but SCIM group provisioning inserted displayName raw. So an IdP that sends "Engineering " creates a team that renders identically to "Engineering" but is a separate row, and membership then splits across the two. The lower(name) unique index from #970 does not catch this, since lower("foo ") and lower("foo") are different strings.

Fix: trim displayName at the read in the three group handlers (POST /Groups, PUT /Groups/:id, and the PATCH replace displayName op), so the dedupe check, the insert, and the response all use the same normalized value the teams API uses. POST already rejected an empty displayName; PUT and PATCH now reject a provided-but-empty name with a 400 (scimError) rather than silently skipping the rename, which would otherwise leave the IdP believing the change applied while the team kept its old name.

Scope: this is the group displayName path the issue reports. userName on the Users paths is left alone, a separate normalization question the issue flagged as check-before-touching.

Testing:
- Added two tests in tests/integration/platform/scim.test.ts: a whitespace twin on POST /Groups now returns 409 against the existing team (it returns 201 without the fix), and a PATCH renaming a group to only whitespace returns 400.
- pnpm --filter ./apps/api typecheck is clean, Biome check is clean, and the SCIM integration test file passes.


thejdubb02:

thejdubb02:

thejdubb02:
recheck

snapotter-hq:
Rebased onto current main, since this had gone conflicting. Worth knowing what I resolved.

Main gained unique-violation handling on the same PATCH displayName path while this sat (#968, then #1153), so the conflict was two real changes landing on one block. Both belong: the resolution rejects the empty name first, then wraps the update in the 409 catch.

I also flipped an older assertion rather than yours. `PATCH replaces displayName and skips empty replacement names` expected a 200 carrying the old name, which is the exact behaviour #988 asks us to stop. The issue's sketch says "reject empty after trim with the existing scimError 400", so that test was stale rather than a disagreement with your change, and it now asserts the 400 and that the name is left alone.

SCIM integration is 88 green. One thing that is not yours: that file has a pre-existing flake where a contender returns 403 from the licence gate instead of racing, so `raceWrites` only ever sees one blocked write. I reproduced it on clean main with no working-tree changes and filed it as #1156.

Thanks for the comment explaining why a silent skip is worse than a rejection. That reasoning is what made the conflict easy to resolve correctly.

thejdubb02:
Thanks for untangling that conflict and flipping the stale 200 assertion, that's the right call per #988. And for chasing down and filing the SCIM race flake separately instead of letting it block this.

Issue #920, opened 2026-08-27
[Bug] Damaged PDF: qpdf timed out after 30s


xromen:
### Before submitting

- [x] I searched existing issues and didn't find a duplicate
- [x] I've tried the latest version to confirm it isn't already fixed

### What happened?

When I try to delete pages in a PDF file larger than 30 megabytes, I get the error "Damaged PDF: qpdf timed out after 30 seconds." Other programs have no problem with this file; the preview works fine for all 90 pages.

### Expected behavior

Pages will be deleted without errors

### Steps to reproduce

1 Upload PDF file to "remove pages"
2 For example type "5,10"
3 Press "remove"

### Affected area

PDF tools

### Specific tool

page remover

### How are you running SnapOtter?

Docker Compose

### Version

 2.2.0

### Deployment details

```shell

```

### Host architecture

AMD64 (Intel/AMD)

### Host OS

Linux (native)

### Logs

```shell

```

PR #994, opened 2026-09-03
fix(pdf): classify a qpdf password-probe timeout as a 4xx, not a 500 (#982)

### LINKED ISSUE #982: qpdf --requires-password timeout during PDF validation surfaces as a 500
### What happens

If `qpdf --requires-password` exceeds its 30s budget during PDF input validation, the raw `QpdfTimeoutError` escapes `validatePdfPath` unclassified. On the HTTP path the tool factory rethrows anything that isn't an `InputValidationError`, so the user gets a generic 500 "Internal server error". On the worker path (ocr-pdf validates inside the process fn) the job fails with the bare reason "qpdf timed out after 30s". Expected: a 4xx with a message that says what actually happened, like the other two qpdf probes now produce.

### Where it lives

- `apps/api/src/modality/document-input.ts`: `validatePdfPath` calls `qpdfRequiresPassword(filePath)` outside any try/catch. The other two qpdf calls in that function are handled since PR #981: a `--check` timeout is treated as inconclusive and skipped, a page-count timeout fails closed with a clear `InputValidationError`.
- `packages/doc-engine/src/qpdf.ts`: `qpdfRequiresPassword` rejects timeouts with `QpdfTimeoutError` (typed in PR #981) after 30s and SIGKILL.
- Escape route: `apps/api/src/routes/tool-factory.ts` (the catch that maps `InputValidationError` to 400 and rethrows everything else) into the global error handler in `apps/api/src/index.ts`.

### Why it wasn't in PR #981

`--requires-password` reads only the encryption dictionary, so it doesn't scale with content the way `--check` does. A 30s stall there means something genuinely anomalous (pathological xref forcing reconstruction, dead disk, crafted file), not "big healthy file on a slow box", so the lenient continue that fixed #920 isn't obviously right. It needs its own decision: fail closed with an honest message, or treat as inconclusive and continue without knowing whether the file is encrypted (which changes which branch of the password policy runs). Fail closed looks safer.

### Starting sketch

Wrap the call using the `isQpdfTimeout` helper that already exists in `document-input.ts`, mirroring the page-count catch from PR #981, and throw an `InputValidationError` along the lines of "Could not inspect this PDF within the time limit". Add a unit test next to "fails closed with a clear error when the page-count probe times out under a page cap" in `tests/unit/api/document-input.test.ts` (mock `qpdfRequiresPassword` to reject with `QpdfTimeoutError`).

### Gotchas

- `qpdfRequiresPassword` cannot throw on nonzero exit: exit 0 means password required, 2 means not encrypted, 3 means encrypted with the right password. Only the timeout path rejects with `QpdfTimeoutError`.
- Related nicety while in there: `runQpdf` collects the last 4 KB of stderr but drops it on timeout. Attaching a tail of it to `QpdfTimeoutError` would make these logs distinguish slow-but-healthy from pathological.

thejdubb02:
Fixes #982.

validatePdfPath (apps/api/src/modality/document-input.ts) called qpdfRequiresPassword outside any try/catch. If the --requires-password probe exceeds its 30s budget, the raw QpdfTimeoutError escaped: on the HTTP path the tool factory rethrows anything that is not an InputValidationError, so the caller got a generic 500; on the worker path the job failed with the bare reason "qpdf timed out after 30s".

The other two qpdf probes in that function are already handled since PR #981: --check treats a timeout as inconclusive and continues (a big-but-healthy PDF on a slow host, issue #920), and the page-count probe fails closed with a clear InputValidationError.

This wraps the --requires-password call the same way, using the existing isQpdfTimeout helper, and fails closed with an InputValidationError. As the issue notes, --requires-password only reads the encryption dictionary, so it does not scale with content the way --check does: a stall there is anomalous (a pathological xref forcing reconstruction, a dead disk, a crafted file), not a healthy file outliving the budget, so failing closed is the safer call. Non-timeout errors are rethrown unchanged. Because validatePdfPath is the shared choke point, both the HTTP and worker paths now get the honest 4xx message.

Testing:
- Added a unit test in tests/unit/api/document-input.test.ts (mocks qpdfRequiresPassword to reject with QpdfTimeoutError): validatePdfPath now rejects with an InputValidationError. It fails without the fix (the raw QpdfTimeoutError escapes).
- pnpm --filter ./apps/api typecheck is clean and Biome check is clean.


thejdubb02:

thejdubb02:

thejdubb02:

thejdubb02:
recheck

thejdubb02:
recheck

snapotter-hq:
Rebased onto #993, so this now carries only the qpdf change instead of showing both.

The distinction you drew is the part that makes it correct: `--requires-password` only reads the encryption dictionary, so it doesn't scale with content the way `--check` does, which is why failing closed here and deliberately skipping on a `--check` timeout (#920) are both right rather than inconsistent. It lines up with what the page-count probe already does, and `InputValidationError` puts it at 400 where it belongs instead of a 500.

Verified against qpdf 12.1.0: document-input unit tests, plus modality-input, remove-pages and sign-pdf on the integration side. 9119 unit tests green.

Thanks for writing the reasoning into the comment instead of just the code.

thejdubb02:
Thanks for rebasing it onto #993 and for laying out why failing closed here and skipping on a --check timeout in #920 are both right. Appreciate the qpdf 12.1.0 verification.

Issue #793, opened 2026-08-14
Landing serves Cloudflare's auto-injected analytics beacon, against the PostHog/Sentry-only egress policy


snapotter-hq:
The landing site serves a third-party analytics script we never added: Cloudflare Pages auto-injects its Web Analytics beacon (`static.cloudflareinsights.com/beacon.min.js`) into every page. TELEMETRY.md says PostHog and Sentry are the only permitted phone-home destinations, and the landing design docs promise near-zero client JS with no third-party egress. This script is both.

## What happens today

Every page load of https://snapotter.com requests `https://static.cloudflareinsights.com/beacon.min.js/...`. Right now the request fails in the browser (`net::ERR_SSL_UNRECOGNIZED_NAME_ALERT`, visible in the console on any visit), so no analytics data actually flows. That failure is luck, not policy: if Cloudflare fixes whatever is wrong with their cert setup, the beacon starts reporting visitor data to Cloudflare silently.

## Where it lives

Not in the repo. The injection is a Cloudflare Pages project setting (Web Analytics auto-injection) on the `snapotter-landing` project. There is nothing to change in `apps/landing/`; grep confirms no reference to cloudflareinsights in the source. The fix is flipping the setting off in the Cloudflare dashboard: Pages project -> Settings -> Web Analytics (or the Cloudflare Web Analytics section) -> disable for this site.

## How it was found

Console audit during the end-to-end verification of the first-party PostHog proxy (#789/#792 follow-up): a headless visit to the deployed landing page logged the failed beacon request. Out of scope for that session because it needs dashboard access, not a code change.

## Fix sketch

1. Disable Web Analytics auto-injection for the Pages project in the Cloudflare dashboard (owner access needed).
2. Reload https://snapotter.com and confirm the console no longer requests `static.cloudflareinsights.com`.
3. Optional hardening so this cannot come back silently: extend `tests/e2e-landing/` with a request-audit spec that fails on any request to a non-allowlisted third-party host, mirroring the app-side egress-allowlist test in `tests/e2e-analytics/egress-allowlist.spec.ts`.

snapotter-hq:
[...]

PR #1051, opened 2026-09-11
fix(web): stop self-hosted instances sending heatmap and dead-click data to PostHog

### LINKED ISSUE #1022: Self-hosted instances send heatmap click data to PostHog because the app never turns capture_heatmaps off
## What happens today

TELEMETRY.md promises that autocapture and session replay are off for the self-hosted app and that every event passes a per-event property allowlist. `apps/web/src/lib/analytics.ts` sets `autocapture: false` and `disable_session_recording: true`, but never sets `capture_heatmaps` (nor `capture_dead_clicks` or `capture_performance`). posthog-js then falls back to the project's server-side toggles, and the SnapOtter project has `heatmaps_opt_in`, `capture_performance_opt_in` and `autocapture_web_vitals_opt_in` all on, because the public sites want them.

Measured on 2026-09-06 with HogQL over the `heatmaps` table, last 7 days: 75,620 click rows from 736 browsers across 2,390 URLs. 13 of those rows come from snapotter.com or docs.snapotter.com; the rest are user instances (sample: `https://ilove.chablais-air.win/`). Each row carries x/y, viewport size, `current_url` and a session id. Web vitals are documented in TELEMETRY.md; heatmaps are not mentioned anywhere.

Expected: an instance sends the allowlisted events plus `$pageview`, `$pageleave` and `$web_vitals`, nothing else. Actual: every click on every instance page is reported with its URL.

## Where it lives

- `apps/web/src/lib/analytics.ts`, the `posthogJs.init` options (around line 70). Heatmap capture is decided by `capture_heatmaps`; when it's unset the SDK asks `/flags` and obeys the project toggle (`heatmaps.ts`, `isEnabled`).
- The `before_send` hook strips query strings from the top-level `$current_url` only. Heatmap data is nested under `$heatmap_data` keyed by full URL, and `$web_vitals` nests `$current_url` inside `$web_vitals_LCP_event` and friends, so both bypass the strip (nested sample seen: `http://<instance>/login`).
- TELEMETRY.md, "SDK-generated events".

## Why it stayed out of #1018 and the public-site analytics PR

Those change the landing and docs, and the project-level heatmap toggle has to stay on for them. The app-side fix touches the privacy invariants, so it wants its own test and its own review.

## Sketch

1. In `initAnalytics`, add `capture_heatmaps: false` and `capture_dead_clicks: false`, and pin `capture_performance: { web_vitals: true, network_timing: false }` so what the doc promises is what the SDK does regardless of project toggles.
2. Extend `before_send` to strip query and fragment from every nested `$current_url` (walk the `$web_vitals_*_event` objects), not only the top-level field.
3. Unit test next to the existing web analytics tests: mock `posthog-js`, call `initAnalytics`, assert the init options carry those keys with those values, so a future SDK default can't reopen this. Those tests run in the node environment, so guard `window` the way `initAnalytics` already does.
4. TELEMETRY.md: list what posthog-js may emit beyond `track()`, and say heatmaps and dead clicks are off.

Prior art: the `$pageview` URL strip in the same `before_send`, and PR #423 for the "opt-out stops all egress" hardening.

thejdubb02:
## What does this PR do?

Stops self-hosted instances from sending heatmap and dead-click data to PostHog, and strips URLs nested in web-vitals events. Fixes #1022.

Fixes #1022

## Type of change

## Checklist

## Details

TELEMETRY.md promises autocapture and session replay are off and only web vitals are captured, but `apps/web/src/lib/analytics.ts` set `autocapture: false` and `disable_session_recording: true` without ever setting `capture_heatmaps`, `capture_dead_clicks`, or `capture_performance`. When those are unset posthog-js obeys the PostHog project's server-side toggles, which stay on for the public sites, so every self-hosted instance reported every click (x/y, viewport, `current_url`, session id).

This sets them explicitly: `capture_heatmaps: false`, `capture_dead_clicks: false`, and `capture_performance: { web_vitals: true, network_timing: false }` so web vitals stay on (as documented) while heatmaps, dead clicks, and network timing do not. It also extends `before_send` to strip the query and fragment from the `$current_url` nested inside `$web_vitals_*_event` objects, which the existing top-level strip missed.

I scoped the nested strip to `$web_vitals_*` per the issue rather than recursively walking every event, to avoid mutating unrelated event structures. Happy to broaden it if you would prefer that.

Tests: two cases in `tests/unit/web/analytics.test.ts`, one asserting the pinned init options and one asserting the nested `$web_vitals` URL is stripped (including multiple and malformed entries). Both fail if the init change is reverted.

### DIFF
--- a/apps/web/src/lib/analytics.ts
+++ b/apps/web/src/lib/analytics.ts
@@ -69,6 +69,15 @@ export async function initAnalytics(config: AnalyticsConfig): Promise<void> {
           api_host: apiHost,
           ...(uiHost ? { ui_host: uiHost } : {}),
           autocapture: false,
+          // Unset, posthog-js obeys the PostHog project's server-side capture
+          // toggles, which stay on for the public sites, so a self-hosted
+          // instance would report every click and dead click (#1022). Set them
+          // explicitly so the client decides: heatmaps and dead clicks off, and
+          // capture_performance limited to web vitals (documented in TELEMETRY.md),
+          // not network timing.
+          capture_heatmaps: false,
+          capture_dead_clicks: false,
+          capture_performance: { web_vitals: true, network_timing: false },
           // Fire $pageview on SPA history changes, not just the initial hard
           // load, so react-router route changes (tool pages, editor, automate,
           // files) are captured. capture_pageleave gives accurate time-on-page.
@@ -89,6 +98,15 @@ export async function initAnalytics(config: AnalyticsConfig): Promise<void> {
               const strip = (u: unknown) => (typeof u === "string" ? u.replace(/[?#].*$/, "") : u);
               props.$current_url = strip(props.$current_url);
               props.$referrer = strip(props.$referrer);
+              // $web_vitals_*_event objects nest their own $current_url, which
+              // skips the top-level strip above, so walk them too (#1022).
+              for (const key of Object.keys(props)) {
+                if (!key.startsWith("$web_vitals_")) continue;
+                const nested = props[key] as Record<string, unknown>;
+                if (nested && typeof nested === "object" && "$current_url" in nested) {
+                  nested.$current_url = strip(nested.$current_url);
+                }
+              }
             }
             return event;
           },
--- a/tests/unit/web/analytics.test.ts
+++ b/tests/unit/web/analytics.test.ts
@@ -287,3 +287,60 @@ describe("analytics lib (baked model)", () => {
     });
   });
 });
+
+describe("privacy invariants (#1022)", () => {
+  it("pins heatmap, dead-click and performance capture so project toggles cannot re-enable them", async () => {
+    await mod.initAnalytics(enabledConfig);
+    expect(mockInit).toHaveBeenCalledWith(
+      "phc_test",
+      expect.objectContaining({
+        capture_heatmaps: false,
+        capture_dead_clicks: false,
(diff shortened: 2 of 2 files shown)


thejdubb02:

thejdubb02:
recheck

snapotter-hq:
Verified on current main: typecheck, lint, 9121 unit tests green.

The mocked `init` means the unit test can only prove what we pass, not that PostHog accepts it, so I checked the three names against the pinned posthog-js (1.404.1). All real init options in `@posthog/types`: `capture_heatmaps`, `capture_dead_clicks`, and `capture_performance`, with `PerformanceCaptureConfig` being exactly `{ network_timing?: boolean; web_vitals?: boolean }`. Since the config goes in as an object literal, excess-property checking means typecheck passing is itself the proof the keys are valid.

Also confirmed the asymmetry holds: `packages/shared/src/analytics/public-site.ts` is untouched and still sets `capture_heatmaps: true`, so the landing and docs keep the behaviour #793 decided on while self-hosted instances stop reporting clicks.

The nested `$web_vitals_*_event` strip is the half I'd have missed. TELEMETRY.md already calls `before_send` the URL boundary for those events, and until this it quietly wasn't one.

Thanks. Merging.

thejdubb02:
Thanks for confirming the public-site asymmetry is intentional and for catching the nested web_vitals strip, that was the part I was least sure I'd gotten right.

Issue #835, opened 2026-08-17
chore(deps): triage the 24 open Dependabot alerts (5 high, pdfjs-dist is runtime-exposed)


snapotter-hq:
Noticed in passing during the security-checklist work (a push printed GitHub's vulnerability banner) and confirmed via the API on 2026-08-17: 24 open Dependabot alerts, 5 high and 19 medium, none critical. The alerts live in the Security tab with full detail; this issue exists so the triage doesn't depend on someone happening to look there. The last dedicated dependency-CVE pass was PR #619 (2026-07-22).

## The five high alerts

| Package | Scope | Patched in | Summary |
|---------|-------|-----------|---------|
| npm/pdfjs-dist | runtime | 6.2.108 | Arbitrary JavaScript execution on opening a malicious PDF |
| npm/fast-uri | runtime | 4.1.2 | Host confusion via backslash authority introducer |
| npm/js-yaml | development | 4.3.1 | Quadratic CPU in !!omap resolution (CVE-2026-59870) |
| npm/undici | development | 7.29.0 | Cross-user info disclosure and parse-time crash via degenerate cache directives |
| npm/undici | development | 8.9.0 | Same advisory, second version range in the lockfile |

All five are lockfile-level (`pnpm-lock.yaml`), so most should resolve with in-range updates rather than manifest changes.

## Priorities and known context

- **pdfjs-dist is the one to treat as urgent.** It's runtime scope and SnapOtter feeds user-supplied PDFs into pdf.js in the web app (the sign-pdf interactive viewer). A malicious uploaded PDF reaching arbitrary JS in the viewer page is exactly our threat model. Check the current version against 6.2.108 and whether the major-version jump breaks the viewer; the sign-pdf e2e specs and the interactive-sign display mode are the regression surface.
- **fast-uri** is runtime via the Fastify chain. In-range bump should do it; the integration suite is the safety net.
- **js-yaml** has a deliberate hold: pinned `^4.2.0` because 5.0.0 is ESM-only and breaks Astro. The patch (4.3.1) is inside the pinned 4.x range, so a lockfile refresh should clear the alert without touching the hold. Verify Astro still builds (`pnpm --filter landing build`).
- **undici** appears twice because two ranges are in the lockfile; both are dev scope.
- The 19 mediums should ride along in the same pass rather than getting individual attention; list them with `gh api "repos/snapotter-hq/SnapOtter/dependabot/alerts?state=open&per_page=100"`.

[...]

PR #1052, opened 2026-09-11
fix(api): validate TRUST_PROXY list at env-parse time instead of crashing Fastify (#999)

### LINKED ISSUE #999: TRUST_PROXY values that are neither boolean nor a valid list crash inside Fastify() with a bare proxy-addr TypeError
A `TRUST_PROXY` value that is neither a boolean literal nor a valid list crashes inside `Fastify()` with a proxy-addr `TypeError` that names neither the variable nor a fix. Found while reviewing the hop-count rejection in #835 (PR #998), which only covers the numeric form.

## Repro

Set any of these and boot the API:

| value | what you get |
| --- | --- |
| `TRUE`, `True`, `yes` | `TypeError: invalid IP address: True` |
| `true ` or `false ` (trailing space, common from a hand-edited `.env`) | `TypeError: invalid IP address: true` |
| `10.0.0.0/8,` (trailing comma) | `TypeError: invalid IP address: ` |
| `10.0.0.0/8;192.168.0.0/16` | `TypeError: invalid IP address: 10.0.0.0/8;192.168.0.0` |
| `0.0.0.0/0` | `TypeError: invalid range on address: 0.0.0.0/0` |

Expected: the process refuses to start at env-parse time with a message that says `TRUST_PROXY=<value>` is wrong and lists the accepted forms, the way the hop-count case does now. Actual: the throw comes out of `Fastify()` in `apps/api/src/index.ts` (around line 344), after migrations, the Redis wait, the SQLite boot import, `INSTANCE_STARTED`, and the OCR reconciliation have all run, as a bare stack trace from fastify's `lib/request.js` (`proxyAddr.compile`). In the all-in-one image s6 restarts the API forever on that. The `true ` case is the confusing one: the operator typed `true` and the log says `invalid IP address: true`.

## Where it lives

`apps/api/src/lib/trust-proxy.ts`, `parseTrustProxy`. The `return value; // CIDR list` branch passes anything through, and the doc comment promises "a comma-separated list of CIDRs or named ranges, all of which proxy-addr understands" without checking. `apps/api/src/lib/env.ts` already calls `parseTrustProxy` from the `superRefine` block (added in #835), so a check placed inside `parseTrustProxy` lands in the right place for free.

Two related misclassifications in the same function, low priority: `127.1` and `2130706433` are IPv4 notations proxy-addr accepts, but `Number()` parses them, so they hit the hop-count message. And `2,10.0.0.0/8` is not a number, so it sails past the guard and proxy-addr trusts `0.0.0.2` next to `10/8`.

## Why it was out of scope

#835 was a dependency sweep; the numeric rejection exists because fastify 5.12.1 changed behaviour under it. This branch was broken the same way before the sweep and is a separate change with its own dependency question below.

## Starting sketch

Validate each comma-separated token in `parseTrustProxy`: accept `loopback`, `linklocal`, `uniquelocal`, a plain IP, or a CIDR, and throw the same style of `TRUST_PROXY=<value>: "<token>" is not ...` message otherwise. Two ways to check a token:

- `ipaddr.js` is already a direct dependency of `apps/api`: `ipaddr.isValid(token)` plus `ipaddr.parseCIDR(token)` in a try/catch. Caveats: `ipaddr.isValid("2")` is true (long-form IPv4 `0.0.0.2`), and `0.0.0.0/0` passes `parseCIDR` but proxy-addr rejects it as a range.
- `@fastify/proxy-addr` `compile()` is what fastify runs anyway, so it is the faithful check, but it is a transitive dependency today and would need adding to `apps/api/package.json`.

Trim tokens before checking so `true ` and `loopback, linklocal` keep working. `tests/unit/security/trust-proxy-resolution.test.ts` has the parser tests and a real-Fastify `resolveIp` helper to extend; `tests/unit/api/lib/env-edge.test.ts` has `expectLoadEnvError` for the env-parse path.

thejdubb02:
## What does this PR do?

Makes a malformed `TRUST_PROXY` fail at env-parse time with an actionable message, instead of crashing inside `Fastify()` with a bare proxy-addr `TypeError`. Fixes #999.

Fixes #999

## Type of change

## Checklist

## Details

`parseTrustProxy` handles `true`, `false`, blank, and the rejected hop-count form, then its final `return value; // CIDR list` branch returned any other string untouched. A value that is neither a boolean nor a valid proxy-addr list (a boolean typo like `TRUE`, `true ` with a trailing space, `10.0.0.0/8,` with a trailing comma, `10.0.0.0/8;192.168.0.0/16` with a semicolon, or a range proxy-addr refuses like `0.0.0.0/0`) therefore reached `Fastify()`. Fastify builds the trustProxy predicate eagerly at construction (`getTrustProxyFn` splits on commas, trims each token, and calls `@fastify/proxy-addr` `compile`), so compile throws `TypeError: invalid IP address: <x>` there. Because `index.ts` constructs Fastify only after migrations, the Redis wait, the SQLite boot and OCR reconciliation, the throw surfaces late as a bare stack trace that names neither `TRUST_PROXY` nor a fix, and in the all-in-one image s6 restarts the API into it forever.

`env.ts` already calls `parseTrustProxy` from a zod `superRefine` (added in #835 for the hop-count case), so a check placed in the CIDR-list branch lands at env-parse time for free. This validates that branch with the same `@fastify/proxy-addr` `compile` Fastify will use, applying Fastify's exact `split(",").map(trim)` first, then throws a message naming `TRUST_PROXY=<value>` and listing the accepted forms, matching the hop-count message. Validating by hand would misjudge shorthands proxy-addr accepts (for example `127.1`), so it reuses proxy-addr rather than reinventing it.

`@fastify/proxy-addr` is added as a direct dependency at `^5.0.0`, which is Fastify's own range, so pnpm dedupes to the single already-installed instance (5.1.0) rather than pulling a second copy. It was previously only transitive via fastify; parse-time validation needs `compile` directly.

I kept this to the crash. The issue notes two low-priority misclassifications in the same function (`127.1` and `2130706433` hit the hop-count message because `Number()` parses them, and `2,10.0.0.0/8` sails past the `Number()` guard so proxy-addr trusts `0.0.0.2` next to `10/8`). Both are unchanged by this PR: the first two are caught by the pre-existing `Number()` guard before the new check runs, and `2,10.0.0.0/8` still compiles in proxy-addr so it passes through exactly as before. Happy to follow up on those separately if you would like.

Tests: two in `tests/unit/security/trust-proxy-resolution.test.ts`. One asserts each malformed value throws the actionable `TRUST_PROXY=<value>` message with the three accepted forms and never leaks the bare `invalid IP address` text; it fails if the fix is reverted. The other pins that `Fastify()` itself throws at construction on those same values, so the reproduction stays honest if a future proxy-addr changes what it rejects.

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -19,6 +19,7 @@
     "@fastify/cors": "^11.3.0",
     "@fastify/http-proxy": "^11.6.0",
     "@fastify/multipart": "^9.0.0",
+    "@fastify/proxy-addr": "^5.0.0",
     "@fastify/rate-limit": "^10.2.0",
     "@fastify/static": "^10.1.2",
     "@neplex/vectorizer": "^0.1.0",
--- a/apps/api/src/lib/trust-proxy.ts
+++ b/apps/api/src/lib/trust-proxy.ts
@@ -9,6 +9,8 @@
  * pin this behaviour without booting the server.
  */
 
+import proxyAddr from "@fastify/proxy-addr";
+
 /**
  * The shipped TRUST_PROXY default: believe `X-Forwarded-For` only from a peer
  * on a private network.
@@ -40,7 +42,8 @@ export const DEFAULT_TRUST_PROXY = "loopback,linklocal,uniquelocal";
  * Two forms, both of which proxy-addr understands:
  *   "true" / "false" -> trust every peer / trust none
  *   anything else    -> a comma-separated list of CIDRs or named ranges
- *                       ("loopback", "linklocal", "uniquelocal")
+ *                       ("loopback", "linklocal", "uniquelocal"), validated
+ *                       here against proxy-addr before it is returned
  *
  * A hop count (a bare number) used to be a third form. fastify 5.12.1
  * (GHSA-3m5p-2c4r-xxw2) stopped honouring it: a hop count cannot validate the
@@ -69,5 +72,22 @@ export function parseTrustProxy(value: string): boolean | string {
         "address.",
     );
   }
-  return value; // CIDR list
+  // Validate the list against the exact parser Fastify will use, so a value
+  // that is neither a boolean nor a valid list fails here, at env-parse time,
+  // instead of throwing a bare `TypeError: invalid IP address: <x>` out of
+  // Fastify() after migrations and Redis are already up (that stack trace names
+  // neither TRUST_PROXY nor a fix, and s6 restarts the API into it forever).
+  // Fastify splits the string on commas and trims each token before compiling
+  // (fastify/lib/request.js getTrustProxyFn), so mirror that split exactly.
+  try {
+    proxyAddr.compile(value.split(",").map((token) => token.trim()));
+  } catch {
+    throw new Error(
+      `TRUST_PROXY=${value}: not a recognized trust value. Use "false" when ` +
+        "nothing proxies this instance, a comma-separated list of CIDRs or named " +
+        `ranges such as "${DEFAULT_TRUST_PROXY}" to name the proxies, or "true" ` +
+        "only when a proxy you control sits in front on a public address.",
+    );
+  }
+  return value; // validated CIDR / named-range list
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -167,6 +167,9 @@ importers:
       '@fastify/multipart':
         specifier: ^9.0.0
         version: 9.4.0
+      '@fastify/proxy-addr':
+        specifier: ^5.0.0
+        version: 5.1.0
       '@fastify/rate-limit':
         specifier: ^10.2.0
         version: 10.3.0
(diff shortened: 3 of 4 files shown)


snapotter-hq:
Verified on top of current main: typecheck, lint, and 9120 unit tests green.

Two things I went looking for. `@fastify/proxy-addr@5.1.0` is already in the tree as fastify 5.12.3's own dependency, so promoting it to a direct dep pulls in no new code at all, and `pnpm install --frozen-lockfile` confirms the lockfile you committed matches. The lockfile diff is three added lines and no changed integrity hashes.

Then the fidelity claim, since a validator that is stricter than the thing it guards would break working deployments on upgrade. Your split mirrors `getTrustProxyFn` in fastify/lib/request.js exactly (`tp.split(',').map(it => it.trim())` then `proxyAddr.compile(values)`), so I ran 25 values through both paths hunting for a case where the guard rejects something Fastify accepts. Zero mismatches: `loopback,linklocal,uniquelocal`, bare IPv6, `fc00::/7`, spaces around tokens, all still fine, and `0.0.0.0/0` was already refused by Fastify before this. Nothing that boots today starts failing.

Thanks for tracing it back to the GHSA and leaving that in the comment. Merging.

thejdubb02:
Thanks for running it against fastify's own getTrustProxyFn and checking the lockfile diff. Good to know the fidelity check came back clean.

PR #1053, opened 2026-09-11
fix(api): log the caught error when an audit write or HMAC step fails (#1012)

### LINKED ISSUE #1012: audit: auditLog drops the error when the DB insert or HMAC step fails, the warn only names the event
Found while reviewing #967. `auditLog` in `apps/api/src/lib/audit.ts` catches both of its failure paths with a bare `catch {}` and warns with only `{ event }`:

```ts
} catch {
  logger.warn({ event }, "Failed to write audit log to DB");
  return;
}
```

and the same shape for "Failed to compute audit HMAC" a few lines below. The error object is dropped, so when rows go missing from `audit_log` the operator gets a line that names the event and nothing else: no Postgres code, no constraint, no message.

## Expected vs actual

Expected: the warn carries `err` so a full disk, a permissions change, a constraint on `audit_log`, or an HMAC key problem can be told apart from the log alone.

Actual: `{ event: "LOGIN_FAILED" } Failed to write audit log to DB`, repeated for every event until someone notices.

## Why it wasn't fixed in #967

That PR added one audit call to the SSO cap denial and stayed on that branch. This is the shared helper every audit call goes through.

## Sketch

```ts
} catch (err) {
  logger.warn({ err, event }, "Failed to write audit log to DB");
  return;
}
```

Same for the HMAC catch. The dual-write design (stdout line first, DB row second) is fine and should stay; the request must not fail because auditing did. Whether a failed audit insert should also reach Sentry through `reportError` (classified operational, so once an hour) is a separate call; the pino `err` serializer alone already fixes the blind spot.

Test: `tests/unit/api/` has no direct `auditLog` unit test today. A container-free one can mock `db.insert` to reject and assert `logger.warn` received the error under `err`.

thejdubb02:
## What does this PR do?

Attaches the caught error to the warning when an audit write fails, so a failure can be diagnosed from the log instead of only naming the event. Fixes #1012.

Fixes #1012

## Type of change

## Checklist

## Details

`auditLog` in `apps/api/src/lib/audit.ts` caught both of its failure paths with a bare `catch {}` and warned with only `{ event }`:

- the DB insert into `audit_log`
- the tamper-resistant HMAC step (settings read, HMAC compute, integrity update)

The error object was dropped, so when rows go missing from `audit_log` the operator gets `{ event: "LOGIN_FAILED" } Failed to write audit log to DB` with no Postgres code, no constraint, and no message: a full disk, a permissions change, and an HMAC key problem all look identical.

This changes both catches to `catch (err)` and adds `err` to the pino context, which is the `{ err, ... }` shape already used elsewhere in the API (for example `enterprise-feature.ts`), so pino's default `err` serializer renders type, message, stack, and enumerable fields like pg's SQLSTATE `code`. It is the fix sketched in the issue.

Control flow is unchanged. The dual-write design stays (stdout line first, DB row second), the insert catch still returns and the HMAC catch still continues, so the request never fails because auditing failed. The only change is that the warning now carries the error.

Tests: `tests/unit/api/audit-mutation.test.ts` already drove both catches ("logs a warning and returns without throwing when the insert fails" and "warns but does not throw when the HMAC update fails") and pinned the warn payload as exactly `{ event }`. Both now expect `{ event, err: expect.any(Error) }`, which keeps the exact-keys guarantee (nothing else leaks in) while asserting the error rides along. Both fail against the pre-fix `catch {}`.

### DIFF
--- a/apps/api/src/lib/audit.ts
+++ b/apps/api/src/lib/audit.ts
@@ -71,8 +71,8 @@ export async function auditLog(
       ipAddress: ip,
       requestId,
     });
-  } catch {
-    logger.warn({ event }, "Failed to write audit log to DB");
+  } catch (err) {
+    logger.warn({ err, event }, "Failed to write audit log to DB");
     return;
   }
 
@@ -100,8 +100,8 @@ export async function auditLog(
         const integrity = computeHmac(rowData, hmacKey);
         await db.update(schema.auditLog).set({ integrity }).where(eq(schema.auditLog.id, id));
       }
-    } catch {
-      logger.warn({ event }, "Failed to compute audit HMAC");
+    } catch (err) {
+      logger.warn({ err, event }, "Failed to compute audit HMAC");
     }
   }
 }
--- a/tests/unit/api/audit-mutation.test.ts
+++ b/tests/unit/api/audit-mutation.test.ts
@@ -365,7 +365,11 @@ describe("auditLog failure handling", () => {
       auditLog(logger as never, "SETTINGS_UPDATED", { userId: "u1" }),
     ).resolves.toBeUndefined();
     expect(logger.warn).toHaveBeenCalledTimes(1);
-    expect(logger.warn.mock.calls[0][0]).toEqual({ event: "SETTINGS_UPDATED" });
+    // #1012: the caught error rides along under `err` (nothing else leaks in).
+    expect(logger.warn.mock.calls[0][0]).toEqual({
+      event: "SETTINGS_UPDATED",
+      err: expect.any(Error),
+    });
   });
 
   it("does not attempt the HMAC branch after an insert failure", async () => {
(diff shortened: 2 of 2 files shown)


snapotter-hq:
Verified on current main: typecheck, lint, 9118 unit tests green.

`err` is the right key rather than an arbitrary one, which is what makes this work: pino's standard serializer picks it up and renders `{type, message, stack}`. I checked it against our actual logger config rather than assuming, and the DB failure now arrives with its stack instead of a bare "Failed to write audit log to DB" and nothing to go on.

Nice touch pinning the log object with `toEqual` instead of `objectContaining`, so the test still fails if something extra ever leaks into that line.

Thanks. Merging.

thejdubb02:
Thanks for double-checking the pino serializer behavior against the real logger config rather than assuming. Glad the toEqual pin on the test held up too.

PR #1121, opened 2026-09-12
fix(search): surface favicon for ICO queries

### LINKED ISSUE #1100: feat: "jpg to ico" finds nothing even though the favicon tool makes ICO
From the in-app search box:

> tambahkan jpg to ico

("add jpg to ico")

## The capability exists under a name nobody searches for

The `favicon` tool (`packages/shared/src/constants.ts:162`, "Generate all favicon and app icon sizes") already produces ICO. Someone who wants a single JPG turned into an `.ico` will not search for "favicon", they will search for "jpg to ico", and that returns nothing.

Same shape as #733, where "edit pdf" failed to surface the pdf-edit tool, and the same shape as #1099 for audio.

## Two ways to fix it, and they are not exclusive

**Keywords.** Add ico and favicon aliases to the `favicon` tool's `keywords` so the existing tool is findable by the words people actually type. Cheapest fix, and it helps immediately.

**A real preset.** `jpg-to-ico` and `png-to-ico` as `BASE_CONFIG` entries. ICO is already a supported output format, so this is a settings lock and a narrowed `sourceInputs`, not new conversion code. Worth it because "one image, one .ico file" is a different job from "generate the full favicon set", and collapsing them into the favicon tool makes people wade through app-icon sizes they did not ask for.

I would do both: keywords now, preset alongside the next conversion batch.

## Watch out for

- ICO holds multiple sizes in one file. A single-image preset should decide whether it writes just the source size or a sensible default set (16/32/48). Silently writing one size when a browser wants several is the kind of thing that generates its own feedback later.
- i18n strings in all 21 locales for any new preset.
- Landing SEO regeneration via `scripts/generate-conversion-seo.mjs`.

## Related

#1099 is the same class of gap for audio. If both get picked up, the shared lesson is that search-miss events are the cheapest source of "the feature exists but nobody can find it" signals, and they are worth reading every release.

dvd233:
## What does this PR do?

Adds ICO-oriented search aliases to the existing favicon tool metadata, so the app and landing-page search surface that tool for `ico`, JPG-to-ICO, and PNG-to-ICO queries (including compact `jpg2ico`/`png2ico` forms).

Fixes #1100

## Type of change

- [ ] New feature (non-breaking change that adds functionality)
- [ ] New translation or i18n update
- [ ] Documentation update
- [ ] Test improvement
- [ ] Refactor (no functional change)

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/snapotter-hq/snapotter/blob/main/CONTRIBUTING.md)
- [ ] I have signed the [CLA](https://github.com/snapotter-hq/snapotter/blob/main/CLA.md) (the bot will prompt you if not)
- [ ] All existing tests pass locally (`pnpm test`)
- [ ] TypeScript compiles without errors (`pnpm typecheck`)

## Validation

- focused Vitest run for `tests/unit/web/fuse-search.test.ts` and `tests/unit/landing/tool-search.test.ts` (60 passed)
- `pnpm exec biome check packages/shared/src/constants.ts tests/unit/web/fuse-search.test.ts tests/unit/landing/tool-search.test.ts`
- `pnpm --filter @snapotter/shared typecheck`
- `pnpm --filter @snapotter/web typecheck`
- `pnpm --filter @snapotter/landing exec tsc --noEmit`

The full repository test and typecheck commands were not run in this sparse local checkout; the affected packages and search suites pass.

## Screenshots (if applicable)

Not applicable; this changes search metadata without changing layout.

### DIFF
--- a/packages/shared/src/constants.ts
+++ b/packages/shared/src/constants.ts
@@ -162,6 +162,7 @@ const BASE_TOOLS: Tool[] = [
     id: "favicon",
     name: "Favicon Generator",
     description: "Generate all favicon and app icon sizes",
+    keywords: ["ico", "jpg to ico", "png to ico"],
     category: "optimization",
     icon: "AppWindow",
     route: "/favicon",
--- a/tests/unit/landing/tool-search.test.ts
+++ b/tests/unit/landing/tool-search.test.ts
@@ -338,6 +338,16 @@ describe("landing searchTools with real catalog metadata", () => {
     expect(result.results[0]?.item.id).toBe("pdf-to-jpg");
   });
 
+  it.each(["ico", "jpg to ico", "jpg2ico", "png to ico", "png2ico"])(
+    "surfaces the favicon tool for %s",
+    (query) => {
+      const result = searchTools(realTools, { query, modality: "all", limit: 5 });
+
+      expect(result.results[0]?.item.id).toBe("favicon");
+      expect(result.hasConfidentMatch).toBe(true);
+    },
+  );
+
   it("does not create confident results for same-format alias conversions", () => {
     const result = searchTools(realTools, { query: "jpg to jpeg", modality: "all", limit: 8 });
 
--- a/tests/unit/web/fuse-search.test.ts
+++ b/tests/unit/web/fuse-search.test.ts
@@ -15,6 +15,11 @@ describe("app fuzzy search finds converters by variation", () => {
     ["heic jpg", "heic-to-jpg"],
     ["mov mp4", "mov-to-mp4"],
     ["convert mp4 to mp3", "mp4-to-mp3"],
+    ["ico", "favicon"],
+    ["jpg to ico", "favicon"],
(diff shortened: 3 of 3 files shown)


dvd233:

dvd233:
recheck

snapotter-hq:
Verified on current main: typecheck, lint, and the full unit suite green at 9128 tests, including your 10 new cases.

I checked the thing that would have made this a bad idea, which is whether the tool you're now pointing "jpg to ico" at actually produces an ICO. It does: `favicon.ts` builds a real multi-size .ico via ImageMagick when it's there, falling back to a 32px PNG, so the search hit isn't sending people to a ZIP of PNGs. `jpg2ico` and `png2ico` pass without being listed because the format-alias generation already covers the `2` forms, which is the right reason for them to work.

Three keywords for a fix that came out of a real `search_miss` event. Thanks. Merging.

Issue #1183, opened 2026-09-17
perf(image): bound the work per-frame animation processing multiplies into


snapotter-hq:
#1083 taught seventeen image tools to keep every frame of an animation. Nine of them do it by running the tool once per frame (`runPerFrame` in `apps/api/src/lib/animated-image.ts`). The loops are all bounded, but the ceilings are high enough to be worth a second look.

## Where the ceiling comes from

`readAnimation` caps the input through `assertGifWorkload`: 500 frames, 16383px per side, 67M pixels across all frames. That bounds what comes in. It does not bound what the work multiplies into.

Three specific ceilings:

**compress in targetSize mode.** `compressToTargetSize` (`packages/image-engine/src/operations/compress.ts`) binary-searches quality (about 7 passes) and can fall into a downscale loop of up to 8 more, each running its own search. Worst case is roughly 56 full animated re-encodes, each touching every frame, on a tool whose `executionHint` is `fast`. Measured on `animated-simpsons.gif` (30 frames, 320x320, ~3 Mpx total): 5.1s when the first pass succeeds. The cap allows about 21x that many pixels before the downscale loop even starts.

**mapFrames retains every transformed frame.** It accumulates each frame as PNG in an array, then hands the whole array to `sharp(frames, { join })`, which decodes all of them together. The cap measures the input; a tool that grows the frame is unbounded from there. Measured: `image-pad` with `{ target: "1:1", padding: 50 }` over 30 x 320x320 took 2492ms and produced 30 x 640x640, four times the area. At the input cap that is roughly 256M pixels of retained frames plus the join's decode.

**image-enhancement with deepEnhance.** The SCUNet call sits inside the per-frame closure, so a 500-frame GIF is up to 500 sidecar invocations and 500 scratch-dir create/remove cycles in one job. The swallowed failure at the `catch` in `image-enhancement.ts` also becomes up to 500 silent failures with nothing logged.

## Why this is a ticket and not a fix

Nothing here is a correctness bug and none of it is reachable by accident: the numbers come from a deliberately large upload. The 202 fallback keeps a long job from failing the request, so the symptom is a worker pinned for a while, not an error. Fixing it properly means deciding on a policy (cap output pixels? cap per-frame work? charge animations a lower frame cap on the expensive tools?) and that decision wants more thought than a bugfix PR should carry.

## Starting sketch

[...]

Issue #1182, opened 2026-09-17
feat(image): let circle-crop and rounded-crop keep animation by emitting WebP


snapotter-hq:
`circle-crop` and `rounded-crop` always encode PNG, so feeding either an animated GIF returns a still of frame 1. #1083 fixed seventeen sibling tools to carry animation; these two are the ones it could not reach.

## Why they were left

Both end with a hard-coded PNG:

- `apps/api/src/routes/tools/circle-crop.ts:96` returns `${base}_circle.png` / `image/png`
- `apps/api/src/routes/tools/rounded-crop.ts:134` returns `${base}_rounded.png` / `image/png`

PNG is the right default for these two, because both punch transparent corners and GIF only has 1-bit alpha, which would give the circle a jagged edge. Sharp cannot write APNG (see #1181), so there is no animated container with good alpha that PNG could simply become.

`tests/unit/api/animated-round-trip.test.ts` pins the current behaviour under "PNG-only tools flatten instead of stacking frames": one frame, at frame size rather than the frames stacked into one tall image. So the flattening is deliberate and guarded, not accidental.

## What would fix it

Switch to animated WebP when the input has more than one frame. WebP has 8-bit alpha, so the circle edge is better than GIF would give, and `mapFrames` in `apps/api/src/lib/animated-image.ts` already knows how to drive a per-frame transform and rejoin with the delays intact. Roughly:

```ts
const animation = await readAnimation(inputBuffer);
const container = animation.animated ? "webp" : undefined;
// still -> png as today; animated -> webp, and the filename extension follows
```

The work is not the encoding, it is the contract change. Output filename and content type stop being constant, so the frontend settings component, the download path and any SEO copy promising a PNG all need to agree. Check `apps/web/src/components/tools/circle-crop-settings.tsx` and its rounded-crop sibling before changing the route.

## Gotchas

- Returning WebP for an animated input and PNG otherwise means the extension is no longer predictable from the tool id. `autoSaveToLibrary` and the ReviewPanel both key off the returned filename.
- Do not reach for GIF as the animated container here. 1-bit alpha is exactly the thing these two tools exist to avoid.

## Out of scope because

#1083 was a bugfix for tools that silently dropped frames. These two do not silently drop anything once its test pins them; changing what they emit is a deliberate format change that deserves its own review.

Issue #1181, opened 2026-09-17
feat(image): carry animation through APNG, which still flattens to one frame


snapotter-hq:
#1083 taught the Sharp-backed image tools to carry animation through GIF and animated WebP. APNG still comes out as a single frame.

## Repro

`tests/fixtures/image/valid/animated.apng` through compress, resize, adjust-colors or any of the other seventeen tools listed in #1083: one frame out, no warning.

## Cause

Two separate walls, and the second is the expensive one.

Detection: Sharp is blind to APNG. `sharp(apngBuffer).metadata()` reports `pages: 1`, so `readAnimation()` in `apps/api/src/lib/animated-image.ts` classifies every APNG as a still and the tools take the single-frame path. The repo already knows this and works around it in `apps/api/src/lib/animation-detect.ts:20`, which routes `.png`/`.apng` through `apngFrameCount()` and reads the `acTL` chunk directly.

Encoding: even with detection fixed there is nowhere to put the frames. Sharp has no APNG writer. `sharp(frames, { join: { animated: true } }).png()` produces a single tall strip, not an animation, and the same is true of any raw rebuild since Sharp ignores `pages`/`pageHeight` on raw input. Verified on sharp 0.35.4 / libvips 8.18.6.

So swapping in `detectAnimation()` alone would make things worse, not better: the tools would start recognising APNG animations they still cannot write, and the likely output is a contact sheet `pages` times too tall.

## Options

1. Convert APNG to animated WebP on the way out and change the extension. Honest, and WebP is the one animated container Sharp writes well, but it silently changes the user's format.
2. Encode APNG ourselves from per-frame PNGs. `apps/api/src/routes/tools/gif-tools.ts:27` already does this trick for GIF (`assembleAnimatedGif` assembles GIF89a at the byte level because Sharp cannot). An APNG equivalent means writing `acTL`/`fcTL`/`fdAT` chunks. Self-contained, no new dependency, and `apngFrameCount` already proves we can parse the format.
3. Reject animated APNG at ingress with a clear message rather than quietly flattening it. Cheap, and better than today's silence.

Option 2 is the real fix. Option 3 is worth doing regardless, since flattening without telling anyone is what made #1083 a user report instead of a known limitation.

## Gotchas

- `mapFrames` in `animated-image.ts` splits and rejoins through PNG already, so a working APNG encoder plugs into the existing seam rather than needing a new one.
[...]

Issue #1180, opened 2026-09-17
fix(duotone): the greyscale step corrupts GIF input because the intermediate re-encodes to a palette


snapotter-hq:
Duotone gives a different answer for the same pixels depending on whether they arrived as a PNG or a GIF.

Build one solid red 100x100 image, encode it both ways, run it through duotone:

```
via PNG -> mean 140,124,87
via GIF -> mean 95,112,92
```

Same pixels, different colour out.

## Cause

`apps/api/src/routes/tools/duotone.ts:39` does the greyscale step with a bare `.toBuffer()`:

```ts
const grayBuf = await sharp(inputBuffer)
  .removeAlpha()
  .grayscale()
  .toColourspace("srgb")
  .toBuffer();
```

No format is named, so Sharp re-encodes in whatever the input was. For a GIF that means writing an image whose interpretation is `b-w` back into a 256-colour palette, and libvips does not handle that combination. A pure red frame comes back as mean 76,105,113, which is not grey at all. `.linear()` then applies the duotone ramp to garbage.

Reproduce the step on its own:

```ts
const step = (b) => sharp(b).removeAlpha().grayscale().toColourspace("srgb").toBuffer();
await step(redAsPng);  // 127,127,127
await step(redAsGif);  // 76,105,113
```

Real photographic GIFs mostly survive it (a simpsons frame greyscales to 161,161,163, close enough to grey to look right), which is why this has gone unnoticed. Flat colour fields are where it falls apart.

## Fix sketch

Name a lossless intermediate instead of inheriting the input's container:

```ts
.toColourspace("srgb")
.png()
.toBuffer();
```

`.raw()` plus an explicit `raw` descriptor on the next `sharp()` would also work and skips an encode. Either way the palette never gets involved.

Worth a scan of the other routes for bare `.toBuffer()` calls on an intermediate, since the same trap is there for any tool that hands a b-w or otherwise unusual colourspace back into a GIF.

## Out of scope because

Found while fixing #1083 (animated frames). That work changed which buffers duotone sees, not what it does to them, so this predates it and is independent of animation: a still GIF hits it too.

PR #1184, opened 2026-09-17
fix(image): keep every frame when tools process animated GIF and WebP

### LINKED ISSUE #1083: fix(compress): animated GIFs come back as a single frame; 22 other image tools share the same sharp() call
Reported through in-app feedback on 2026-08-14 (app 2.2.0), against `compress`:

> mon gif est devenue une image statique

("My gif became a static image.")

Confirmed. Compressing an animated GIF returns frame 1 and drops the rest.

## Cause

`apps/api/src/routes/tools/compress.ts:19` builds the pipeline without the animated flag:

```ts
const image = sharp(inputBuffer);
```

Sharp decodes only the first frame unless you pass `{ animated: true }`. Everything downstream sees a single-frame image, so the re-encode writes a single-frame GIF.

## Repro

Using the fixtures already in the repo:

```
tests/fixtures/image/valid/animated.gif           3 frames  -> 1 frame
tests/fixtures/image/valid/animated-simpsons.gif  30 frames -> 1 frame
```

```js
await sharp(buf).gif().toBuffer();                  // pages = 1
await sharp(buf, { animated: true }).gif().toBuffer(); // pages = 3 / 30
```

Expected: an animated GIF in, an animated GIF out, smaller. Actual: a still frame, and the size drop looks like a successful compression so nothing warns the user.

## Prior art

The GIF-aware tools already do this correctly, so the pattern to copy is in the repo:

- `apps/api/src/lib/animation-detect.ts:24`
- `apps/api/src/routes/tools/gif-tools.ts:106`
- `apps/api/src/routes/tools/gif-webp.ts:26`

## Blast radius

23 tool routes construct sharp the same way (`grep -rln "sharp(inputBuffer)" apps/api/src/routes/tools/`):

`pixelate, text-overlay, replace-color, lqip-placeholder, rotate, smart-crop, vignette, adjust-colors, compress, resize, split, optimize-for-web, image-enhancement, crop, watermark-text, color-blindness, image-pad, duotone, histogram, circle-crop, rounded-crop, sharpening, strip-metadata`

They don't all need fixing. `histogram`, `lqip-placeholder`, `smart-crop` and `split` reduce to a single frame by definition. But `compress`, `optimize-for-web`, `resize`, `rotate`, `crop`, `strip-metadata`, `watermark-text` and the colour adjustments should all round-trip animation, and today none of them do. Worth auditing the list against `acceptedInputs` in `packages/shared/src/constants.ts` to see which actually accept `.gif`, `.webp` or `.apng` and fixing that subset.

## Gotchas

- `{ animated: true }` lays frames out as one tall strip, so height is `pageHeight`, not `height`. Any operation doing its own geometry has to use `pageHeight` or it will distort every frame. This is the reason to fix per-tool rather than sed the whole list.
- Animated WebP and APNG go through the same decode path, so the fix covers them too.
- Frame count is unbounded. `gif-tools.ts` already caps frames; whatever lands here should reuse that cap rather than invent a second one.

## Out of scope because

Found while mining PostHog feedback for landing-page quotes.

## Starting sketch

For `compress` alone, detect animation first and only then opt in, so still images keep the current single-frame path:

```ts
const animated = await isAnimated(inputBuffer); // lib/animation-detect.ts
const image = sharp(inputBuffer, animated ? { animated: true } : undefined);
```

Then a matrix test over the two animated fixtures asserting `metadata().pages` survives the round trip for each tool claimed to support animation.

snapotter-hq:
`mon gif est devenue une image statique`. Compressing an animated GIF gave back frame 1 and threw the rest away, at a byte count small enough to look like the compression had worked. Sixteen other tools had the same bug, all from one line: `sharp(inputBuffer)` decodes a single frame unless you ask for more.

## Why the one-line version does not work

`{ animated: true }` keeps the frames but stacks them into one tall strip, and Sharp's operations split four ways rather than behaving uniformly:

- `resize`, `extract` and `extend` are page-aware and need nothing.
- `composite` paints the strip, so an overlay lands on frame 1 alone. `flip` mirrors the strip, which reverses the frame order on top of flipping each frame.
- `sharpen`, `blur`, `median` and `convolve` sample neighbouring rows, so at the seam they read the next frame. Measured on two flat frames (100 and 200), `sharpen({sigma: 3})` turns frame 0's last rows into 70, 54, 54 and frame 1's first rows into 228.
- `rotate` refuses multi-page input outright, and anything rebuilt from `.raw()` comes back single-page because Sharp ignores `pages`/`pageHeight` on raw input.

The nastiest is encoding a strip to a still format. It does not drop the extra frames, it writes them: a 100x100 GIF converted to JPEG comes out 100x300.

## What changed

`apps/api/src/lib/animated-image.ts` holds two paths, and the choice per tool is about which group above its operations fall into.

`readAnimationFor(buffer, outputFormat)` reports frames only when the output container can hold them, so conversions to JPEG, PNG and AVIF stay on the single-frame path, which is the right answer for them anyway. Multi-frame input goes through the existing `assertGifWorkload` cap instead of a second limit.

Eight tools open the strip in one pass: adjust-colors, color-blindness, compress, crop, optimize-for-web, resize, strip-metadata, and adjust-colors again for everything but its sharpness setting.

Nine go frame by frame through `runPerFrame`: duotone, image-enhancement, image-pad, pixelate, replace-color, rotate, sharpening, text-overlay, vignette, watermark-text. It splits to PNG, hands each frame to the tool's existing logic unchanged, and rejoins with per-frame delays and the loop count restored. Frames travel as PNG both ways so nothing is lost on the way in.

Four more bugs surfaced while verifying this and are fixed here, because each one drops or mangles frames through the same code path:

- `resize` and `crop` in the image engine sized themselves off `metadata.height`, which is the strip. A 50% resize of a 100x100x3 GIF produced 100x300 frames; an out-of-frame crop slipped past the bounds check and died inside libvips instead of returning `Crop region exceeds image height`.
- `compressToTargetSize` re-opens the buffer on every search pass, so the flag had to reach it. `CompressOptions` gained `animated`.
- `replace-color` with `makeTransparent` and `image-pad` with a transparent background both fell back to PNG, which cannot hold frames. GIF has 1-bit alpha so the container does have to change; it is now WebP, which has both alpha and frames.
- `image-enhancement` ran its auto-exposure analysis per frame, so a darker frame got a lift its neighbour did not (measured: brightness +10 on one frame, -12 on the next). It now measures once per upload.

`circle-crop` and `rounded-crop` always encode PNG, which Sharp cannot animate, so they still return a still. That is now pinned by a test rather than being accidental, and #1182 tracks moving them to WebP.

## Verified

- `tests/unit/api/animated-round-trip.test.ts`, 72 cases: frame count and per-frame geometry for 17 tools across a GIF and a WebP, delays and loop count, rotate at a real angle, frame order under vertical flip, seam bleed on the neighbourhood operators, still-format conversions, and the PNG-only pair.
- Teeth: reverting the fix fails 38 of them. Removing the output-format gate fails the 3 still-format cases. Dropping the delay and loop arguments fails exactly the 9 per-frame tools.
- `pnpm lint` and `pnpm typecheck` clean. `pnpm vitest run tests/unit/api tests/unit/image-engine`: 4166 passed.
- `pnpm test` full suite: 20423 passed, 1 failed. Baseline on main at `0f4857af` was 20351 passed, 1 failed. The one failure is the known SCIM concurrent-rename race (#980, #990, #1000, #1156), which fails on a different case each run and touches nothing this branch changes. Zero new failures.
- Runtime is unchanged: 980.7s on the baseline, 977.1s here.

## Tradeoff worth knowing

Per-frame processing multiplies work by the frame count. Measured on a 30-frame 320x320 GIF: the one-pass tools stay under 250ms, the per-frame ones run 1.5 to 1.9 seconds. That is well inside the sync window, and a large enough animation falls back to 202 the way any slow job does. The ceilings this creates are written up in #1183 rather than guessed at here.

Follow-ups filed: #1180 (duotone corrupts GIF input through a palette re-encode, predates this), #1181 (APNG still flattens, Sharp has no APNG writer), #1182 (circle-crop and rounded-crop), #1183 (per-frame resource ceilings).

Closes #1083


snapotter-hq:
Landing the pr-review-toolkit output here so it does not stay in a session transcript. Three agents ran on this diff: silent-failure-hunter, code-reviewer, pr-test-analyzer. They found four real bugs this PR had introduced, all fixed before merge.

## Found and fixed in this PR

**`resize` and `crop` sized themselves off the frame strip.** The engine reads `metadata.height` off the pipeline, which under `animated: true` is `pageHeight * pages`. Sharp then applies that per page, so every frame came out three times too tall on a 3-frame GIF:

```
resize {percentage:50}                        -> pages=3 width=50  pageHeight=150   want 50x50
resize {width:500,height:500,withoutEnlargement} -> pages=3 width=100 pageHeight=300   want 100x100
crop  {unit:"percent"}                        -> THREW extract_area: bad extract area
crop  {left:0,top:0,width:100,height:250}     -> THREW, bypassing the bounds guard
```

That last one is the worst: the guard that exists to return `Crop region exceeds image height` measured against the strip, so an out-of-frame crop sailed past it into a raw libvips error reported as a server fault. Both engine operations now read `pageHeight ?? height`.

My test missed all of this because `defaultSettingsFor` supplies `resize: { width: 64 }` and a pixel-unit in-bounds crop, which are exactly the two settings that dodge it. Explicit cases added for the dangerous ones.

**`sharpen`, `blur`, `median` and `convolve` read across the frame seam.** My three-way taxonomy of Sharp operations was missing a fourth group. These sample neighbouring rows, so on a stacked strip each frame picks up a halo from its neighbour. Measured on two flat frames (100 and 200), `sharpen({sigma:3})` turned frame 0's last rows into 70,54,54 and frame 1's first rows into 228. The frame count stays correct, which is what makes it easy to miss. `sharpening` moved to the per-frame path, and `adjust-colors` routes there only when `sharpness > 0`.

**`replace-color` with `makeTransparent` and `image-pad` with a transparent background still flattened.** Both force PNG when a setting needs alpha, and PNG cannot hold frames, so the animation was silently discarded in the exact way this PR exists to stop. Both now choose WebP for animated input, which has both alpha and frames.

**`runPerFrame` applied the frame cap before the container gate**, so a 501-frame GIF headed for PNG output was rejected over a frame count that was about to be discarded, while the same input through `openAnimated` returned 200. Reordered to match `readAnimationFor`.

**`readAnimation` swallowed a metadata failure silently**, which would quietly reinstate the frame dropping this PR fixes. Now logged.

**Smaller ones:** the frame cap said "GIF input" for animated WebP; `mapFrames` handed each frame to the tool through a lossy encode (mean 5.71 per channel on WebP) and now uses PNG both ways; `rotate` rebuilt the animation even for a no-op; a docstring described the frame container backwards; `AnimationInfo` collided with the existing export in `apps/api/src/lib/animation-detect.ts` and was renamed `FrameInfo`; `isAnimated` was awaited on paths that discard it; two `--` stand-ins sat on lines this diff rewrites.

**`image-enhancement` measured auto-exposure per frame.** Verified: frame 0 got brightness +10 and frame 1 got -12, so each frame was corrected against itself, which plays back as flicker. Now measured once per upload.

**Test gaps closed:** nothing asserted frame delay or loop count (dropping them leaves every frame-count assertion green, and rebuilding without them gives `delay [0,0,0], loop 1`); the geometry cases passed vacuously when everything was flattened and are now folded into the frame-count assertion; the fixture is built rather than read, with brightness-distinct frames and non-default timing, because the flat-colour `animated.gif` makes several tools produce identical frames that the encoder then legitimately merges.

## Considered and deliberately not done

**A regression test for the 500-frame cap.** I could not build a reliable 501-frame fixture: the encoder merges identical consecutive frames and small flat frames quantise into each other, so 501 distinct 8x8 frames came back as 63 pages. Noted in #1183 rather than faked.

**A regression test for the image-enhancement flicker fix.** I wrote one, measured it, and it could not tell fixed from broken (a 1-unit difference). Deleted rather than ship a test that passes either way. The bug itself is verified by the +10 and -12 measurement above.

**Per-frame processing multiplies work by frame count.** Measured on a 30-frame 320x320 GIF: one-pass tools stay under 250ms, per-frame tools run 1.5 to 1.9 seconds. Documented as a tradeoff in the PR body; the ceilings it creates are #1183.

**Two dead exports** (`tilePerFrame`, `tileAllPerFrame`) left over from the approach I abandoned were removed, and `readAnimation`, `mapFrames` and `animatedContainerFor` were unexported, so the public surface is only what routes use.

## Filed instead

#1180, #1181, #1182, #1183, and later #1185 from the nightly.

Issue #1190, opened 2026-09-19
fix(image): unnamed intermediate buffers re-encode in the input's container, corrupting still GIFs


snapotter-hq:
Several image tools write an intermediate buffer with a bare `toBuffer()`. Sharp then re-encodes that intermediate in whatever container the input arrived in, which inserts a lossy generation the user never asked for. #1180 fixed one instance of this in duotone, where the symptom was a GIF palette; the more common symptom is a JPEG re-encode at quality 80.

## The JPEG case

`resolveOutputFormat` asks for quality 95, but an unnamed intermediate uses Sharp's JPEG default of 80. Measured through `circle-crop`'s extract step on a 200x200 high-frequency pattern encoded at quality 95:

```
intermediate format when unnamed: jpeg
RMSE added by the unnamed intermediate vs .png(): 22.48
```

Photographs will sit well under that, since the pattern is a worst case, but the mechanism is an extra lossy round trip inserted into the middle of a pipeline the user asked to run at 95.

## Where

Scanned `apps/api/src/routes/tools/` for `const x = await sharp(...)...toBuffer()` chains that name no encoder:

- `apps/api/src/routes/tools/circle-crop.ts:55`
- `apps/api/src/routes/tools/rounded-crop.ts:94`
- `apps/api/src/routes/tools/smart-crop.ts:78`, `:143`, `:152`
- `apps/api/src/routes/tools/stitch.ts:328`, `:333`, `:341`, `:366`, `:371`, `:379`, `:406`, `:411`, `:418`
- `apps/api/src/routes/tools/collage.ts:654`
- `apps/api/src/routes/tools/passport-photo.ts:88`
- `apps/api/src/routes/tools/vignette.ts:68`

`apps/api/src/routes/tools/stitch.ts` is the worst of them: nine sites on the resize path, one per input image, so a stitch of eight JPEGs pays eight unrequested re-encodes.

Two that the same scan flags and that are fine as they are:

- `apps/api/src/routes/tools/vignette.ts:66` reads an SVG, and Sharp falls back to PNG for SVG input, so that intermediate is already lossless.
- `apps/api/src/routes/tools/duotone.ts:61` follows a step that now names PNG, so its output is PNG regardless. Naming it again would be a no-op, which is why #1180 left it alone.

`apps/api/src/routes/tools/image-enhancement.ts:76` is the same pattern with a worse symptom and is tracked separately in #1187.

## Fix sketch

Name a lossless intermediate at each site:

[...]

Issue #1189, opened 2026-09-19
fix(duotone): the intensity blend reads past the end of its buffer silently if the two raw sides ever disagree


snapotter-hq:
The intensity blend in duotone trusts that two independently-built raw buffers have the same length. They do today. If that ever stops being true, the tool returns a partly black image with a 200 and nothing is logged.

## The code

`apps/api/src/routes/tools/duotone.ts`:

```ts
const origRaw = await sharp(source).removeAlpha().toColourspace("srgb").raw().toBuffer({ resolveWithObject: true });
const duo = await sharp(buf).removeAlpha().toColourspace("srgb").raw().toBuffer();
const pixels = origRaw.data;
const blended = Buffer.alloc(pixels.length);
for (let i = 0; i < pixels.length; i++) {
  blended[i] = Math.round(pixels[i] * (1 - k) + duo[i] * k);
}
```

The loop runs to `pixels.length` and indexes `duo[i]`. The invariant that the two match is stated in a comment above it, which is the only thing holding it.

## Why it is currently safe

`removeAlpha().toColourspace("srgb")` normalises everything. Checked against 16-bit RGB PNG, one-channel greyscale, greyscale plus alpha, CMYK JPEG, CMYK TIFF, palette GIF, transparent GIF, animated GIF, and images built with 6, 9, 12 and 18 bands: every one comes back as exactly 3 channels of `uchar`. Both sides derive from the same source dimensions, so the lengths agree.

## Why it is worth a guard anyway

The failure mode is silent, not loud. `Math.round(undefined)` is `NaN`, and assigning `NaN` to a Buffer index writes 0 without throwing. Separately, `sharp(buffer, { raw })` only rejects a buffer that is too *short*; an over-long one is accepted.

Measured, by dropping `removeAlpha()` from the original side only on a 60x60 input:

```
NaN bytes: 3600 of 14400, all written as 0
rewrap: succeeded, job reported success
the bottom 15 rows of the output are pure black
```

So a plausible future edit to one of the two pipelines, exactly the kind of edit #1188 will require when it re-attaches the alpha channel, turns part of every image black with no error anywhere.

## Fix sketch

```ts
if (duo.length !== pixels.length) {
  throw new ToolInputError(
    `duotone blend: duotone buffer is ${duo.length} bytes, original is ${pixels.length} ` +
      `(${origRaw.info.width}x${origRaw.info.height}x${origRaw.info.channels})`,
  );
}
```

[...]

Issue #1188, opened 2026-09-19
fix(duotone): a transparent image comes back with an opaque shadow-coloured background


snapotter-hq:
Duotone a logo with a transparent background and the transparency is replaced by a flat rectangle in the shadow colour.

## Repro

A red disc on a transparent ground, 80x80 PNG, through the real route:

```
input:            corner=rgba(0,0,0,0)      centre=rgba(200,50,50,255)   hasAlpha=true
intensity 100 ->  corner=rgba(30,58,138,255) centre=rgba(122,113,95,255)  hasAlpha=false
intensity 0   ->  corner=rgba(0,0,0,255)     centre=rgba(200,50,50,255)   hasAlpha=false
```

`30,58,138` is `#1e3a8a`, the default shadow colour. The transparent ground became an opaque rectangle painted in it. Same for a transparent GIF or WebP. Response is 200, the filename keeps its usual `_duotone` suffix, and nothing in the payload says the alpha channel is gone.

The intensity 0 row is the sharper one. The slider reads as "no effect" in `apps/web/src/components/tools/duotone-settings.tsx`, and the colours are indeed untouched, but the file still comes back without its alpha channel and with a black background. Someone nudging the slider to zero to compare gets a different image than they put in.

## Cause

`apps/api/src/routes/tools/duotone.ts` strips alpha on both pipelines and never puts it back:

```ts
const grayBuf = await sharp(source).removeAlpha().grayscale()...   // line 51
const origRaw = await sharp(source).removeAlpha().toColourspace("srgb").raw()...  // line 65
```

`removeAlpha()` is needed: the duotone ramp is applied with `.linear()` over three channels, and a fourth band would be transformed along with them. The gap is that nothing re-attaches the band afterwards.

## Fix sketch

Hold the alpha band aside and rejoin it after the ramp:

```ts
const meta = await sharp(source).metadata();
const alpha = meta.hasAlpha
  ? await sharp(source).extractChannel(3).png().toBuffer()
  : undefined;
// ... existing pipeline, unchanged ...
if (alpha) buf = await sharp(buf).joinChannel(alpha).png().toBuffer();
```

Name the intermediate format explicitly. A bare `toBuffer()` there re-encodes the one-band mask in the input's own container, which is #1180 all over again and is exactly the defect #1187 describes in image-enhancement.

[...]

Issue #1187, opened 2026-09-19
fix(image-enhancement): a transparent GIF comes back opaque because the alpha mask re-encodes to a palette


snapotter-hq:
Run a GIF with a transparent region through image-enhancement and the transparency is gone. The same pixels as a PNG keep it.

## Repro

A 60x60 red square with a fully transparent 20x20 corner, encoded both ways, through the real route:

```
PNG input: alpha at (5,5)=0 [transparent], (40,40)=255 [opaque]
  enhanced -> image/png  alpha (5,5)=0    (40,40)=255      correct
GIF input: alpha at (5,5)=0 [transparent], (40,40)=255 [opaque]
  enhanced -> image/gif  alpha (5,5)=255  (40,40)=255      transparency lost
```

## Cause

`apps/api/src/routes/tools/image-enhancement.ts:78` lifts the alpha channel out so the corrections can run on colour only, then rejoins it afterwards:

```ts
let alphaBuffer: Buffer | undefined;
if (hasAlpha) {
  alphaBuffer = await sharp(inputBuffer).extractChannel(3).toBuffer();
}
```

`extractChannel(3)` leaves a single-band mask, and the bare `.toBuffer()` names no format, so Sharp re-encodes it in whatever the input was. For a GIF that means writing a one-band mask through a 256-colour palette, and it comes back as something else entirely:

```
PNG in -> extractChannel(3).toBuffer() -> fmt=png ch=1 mean=255   the mask
GIF in -> extractChannel(3).toBuffer() -> fmt=gif ch=4 mean=76    not the mask
```

The `joinChannel(alphaBuffer)` at line 96 then rejoins that, and the result no longer describes the original transparency.

This is the same trap as #1180, which fixed two bare `.toBuffer()` calls in duotone. I scanned for it while fixing that one; this is the only other instance I found where a chain changes the interpretation of the image and then writes an intermediate without naming a format.

## Fix sketch

Name a lossless intermediate, the way #1180 did:

```ts
alphaBuffer = await sharp(inputBuffer).extractChannel(3).png().toBuffer();
```

`.raw()` would also work and skips an encode, but it needs care: a one-band raw buffer has to be rewrapped with an explicit `raw` descriptor of `channels: 1`, and reading it back through `sharp()` without that returns three channels. That combination already caused #1082, so PNG is the safer choice here unless the extra encode shows up in a profile.

## Gotchas

[...]

PR #1193, opened 2026-09-19
fix(duotone): name the grayscale intermediate so GIF input keeps its luminance

### LINKED ISSUE #1180: fix(duotone): the greyscale step corrupts GIF input because the intermediate re-encodes to a palette

snapotter-hq:
Duotone answered by container rather than by pixels. A flat red square as a PNG came out `140,124,87`. The same pixels as a GIF came out `76,105,113`, and so did green, and so did blue: every saturated color collapsed onto one value.

## Why

The grayscale ramp went through an intermediate `toBuffer()` that named no format, so Sharp re-encoded it in whatever the input was.

Sharp's GIF writer defaults to `reuse: true`, re-expressing the image in the palette libvips attached when it read the source. So a grayscale ramp has to be built out of the colors the source happened to contain. A flat red GIF has no grays in its palette at all, and the result bears no relation to the input. A GIF whose palette already holds grays round-trips exactly:

```
gray-palette GIF  90,90,90   -> 90,90,90        fine
flat red GIF      200,0,0    -> 76,105,113      collapsed
photographic GIF             -> 101,97,101      vs 100,100,100 as PNG
```

That last row is why nobody reported it: a photograph's 256-entry palette has near-grays throughout, so it drifted instead of collapsing. Drifting is not nothing, though. Per pixel, a photographic GIF was off by up to 81 while the channel means barely moved.

## What changed

One `.png()` on the grayscale intermediate.

The second intermediate is deliberately left alone. Once the first names PNG, `sharp(grayBuf).linear(...).toBuffer()` already emits PNG, so naming it again is a no-op that no test could fail on. I confirmed all four combinations rather than assuming:

```
both .png()      -> 140,124,87    correct
only 1st .png()  -> 140,124,87    identical
only 2nd .png()  -> 95,112,92     still wrong
neither          -> 76,105,113    the bug
```

The comment records that the chain depends on the first call, since that dependency is otherwise action at a distance.

Animated GIFs were never affected: since #1083 they are split into PNG frames before the tool runs, so this only ever hit stills. That is also why the animated round-trip suite stayed green through it.

## Verified

- `tests/unit/api/duotone-container-independence.test.ts`, 11 cases: GIF, WebP and JPEG agreeing with PNG for three colors at two intensities; three colors through a GIF staying distinct; a flat color landing where the ramp arithmetic independently says; and a photographic GIF matching its own PNG transcode pixel for pixel.
- Teeth: reverting the fix fails all 11, restoring passes all 11. The photographic cases fail too, which is what confirms photos were affected.
- Intensity 40 rather than 50 for the blend branch. At 50 the blend's two weights are equal, so the one case meant to exercise it would not notice them being inverted.
- `pnpm lint` exit 0, `pnpm typecheck` 9 of 9, `pnpm vitest run tests/unit/api tests/unit/image-engine` 4176 passed, `tests/integration/tools/image/duotone.test.ts` 4 passed, `animated-round-trip.test.ts` 72 passed.
- Full `pnpm test`: 20434 passed, 1 failed. Baseline on main at `1e524902` was 20423 passed, 1 failed. Same failure both sides, the known SCIM concurrent-rename race (#980, #990, #1000, #1156), which touches nothing here. Zero new failures.

One earlier run of the full suite came back with 884 failed files. That was the Docker VM restarting mid-run and taking Redis with it, not this change: every container on the host, including unrelated ones, had restarted. Re-run on healthy infra gave the numbers above.

## Tradeoff

PNG intermediates cost more to encode for some inputs. Measured end to end on a 12MP photograph: JPEG input goes from 417ms to 878ms, WebP input goes the other way, from 4004ms to 1682ms, and PNG input is unchanged. Intermediate bytes for a JPEG input go from 1.7MB to 17MB. Net it looks like a fair trade for a pipeline whose whole job is a luminance ramp, and it removes two lossy generations that were never asked for. `.png({ compressionLevel: 1 })` would claw back most of the JPEG cost if it ever matters; there is no evidence it does, so it stays a note rather than a change.

## Follow-ups filed

The issue asked for a scan of other routes for the same trap. It found more than expected:

- #1187: image-enhancement lifts the alpha channel out through the same unnamed intermediate, so a transparent GIF comes back opaque.
- #1188: duotone strips alpha and never restores it, so a transparent logo gets an opaque background in the shadow color. Separate defect in the same file.
- #1189: the intensity blend reads past the end of its buffer if the two raw sides ever disagree, writing zeros rather than failing. Unreachable today, which is why it is not fixed here: there is no failing test to justify the code.
- #1190: the same unnamed-intermediate pattern across other tools. On still GIFs it corrupts rather than degrades, verified on `border` (a green border comes out `76,105,113`) and `vectorize`. My first scan of this undercounted badly, so the issue carries a correction and a suggested catalogue-wide guard.

Closes #1180


snapotter-hq:
Landing the pr-review-toolkit output here so it does not stay in a session transcript. Three agents ran on this diff: code-reviewer, pr-test-analyzer, silent-failure-hunter.

## Found and fixed in this PR

**The second `.png()` was a no-op.** Both agents caught it independently. Once the first intermediate names PNG, `sharp(grayBuf).linear(...).toBuffer()` already emits PNG. I checked all four combinations rather than taking it on trust:

```
both .png()      -> 140,124,87    correct
only 1st .png()  -> 140,124,87    identical
only 2nd .png()  -> 95,112,92     still wrong
neither          -> 76,105,113    the bug
```

Removed it, so every line in the diff is load-bearing and covered.

**My explanation of the cause was wrong.** I had written "a 256-colour palette", implying quantisation. The real mechanism is that Sharp's `.gif()` defaults to `reuse: true` and re-expresses the image in the palette libvips attached when it read the source. Verified: a gray-palette GIF round-trips exactly (90 to 90), a flat red one collapses. That is also what explains why photographs survived, which the old wording did not. Both the code comment and the test docblock were rewritten.

**Intensity 50 was the worst possible choice for the blend case.** At `k = 0.5` the blend's two weights are equal, so inverting them changes nothing and the one case meant to exercise that branch would not notice. Moved to 40.

**A hard-coded luminance of 127 pinned a libvips constant, not our behaviour.** Now derived at runtime. Doing that turned up a trap worth recording: `sharp().stats()` reports on the image as loaded and ignores queued operations, so `sharp(buf).grayscale().stats()` returned the original red channel (255) and silently made the expected value the highlight colour. The grayscale has to be written out before it is measured.

**No absolute anchor on the blend branch.** Every cross-container case compared against PNG computed by the same code, so a regression breaking all containers equally would have passed. Added the ramp arithmetic at both intensities.

**No photographic coverage, and the obvious metric would have hidden it.** Photographic GIFs drifted rather than collapsing: up to 81 per pixel while the channel means barely moved. A means-based photographic case would have passed with the bug present. Added a per-pixel comparison against the file's own PNG transcode.

**GIF decodes to four channels and PNG to three**, so the first version of that comparison failed on buffer length rather than on pixels. Both sides now go through `ensureAlpha()`.

**British spelling.** The new prose used "greyscale" and "colour" in a codebase that is 189 to 6 and 1843 to 2 the other way, with `apps/api/src/routes/tools/duotone.ts` saying "Grayscale" and "greyscale" five lines apart. Normalised, leaving Sharp's own `toColourspace` API alone.

## Considered and deliberately not done

**Naming `.png()` on the second intermediate anyway, for robustness.** code-reviewer argued the current correctness is action at a distance from the first call. Declined: it is a provable no-op, so no test can fail on it, and it would not protect anything either, since naming only the second still returns the wrong answer (95,112,92 above). The dependency is recorded in the comment instead.

**The blend-branch length guard.** silent-failure-hunter is right that `duo[i]` returning undefined would write zeros and hand back a partly black image with a 200. I could not construct any input that reaches it: `removeAlpha().toColourspace("srgb")` normalises 16-bit, 1-channel, CMYK, palette and multi-band input to exactly 3 channels. Adding untestable defensive code did not belong in a bugfix PR, so it is #1189 with the measurements.

**Removing the `as 1 | 2 | 3 | 4` cast.** Tied to the guard above, deferred with it.

**PNG intermediates cost more for some inputs.** Measured on a 12MP photograph: JPEG input 417ms to 878ms, WebP input 4004ms to 1682ms, PNG unchanged. Per the repo rule on speculative optimisation this stays a documented tradeoff in the PR body rather than an issue, since nothing points at it mattering.

**Extracting `swatch()` and `channelMeans()` into `tests/helpers/`.** No other route has this test shape yet. Premature until #1190 builds the catalogue-wide version.

**Adding an integration-level GIF case.** The unit test already exercises production's path: GIF is in `IMAGE_INPUTS` and not in `CLI_DECODED_FORMATS`, so the factory hands `process` the original bytes. #1190 proposes the durable version.

**16-bit input is silently downcast to 8-bit** by `toColourspace("srgb")`. Defensible for a two-color ramp whose output has at most 256 luminance steps by construction, pre-existing, and not worth a ticket on its own.

## Filed instead

#1187, #1188, #1189, #1190.

Issue #1201, opened 2026-09-19
docs(chart-maker): the tool page still documents the two-column contract that was removed


snapotter-hq:
`apps/docs/tools/files/chart-maker.md:47` still documents the contract #1158 removed:

> The first column is used as the category label; the second column must be numeric and provides the data values. Only two columns are used.

All three of those Notes bullets are now wrong. Column detection replaced the fixed pair, JSON takes arbitrary keys and a single-array wrapper, and dispatch reads the content instead of the file extension.

What the page should say:

- The value column is the rightmost one holding numbers in more than half the rows. The label column is the leftmost one that doesn't.
- If every column is numeric, the first is the label and the last is the value.
- A single-column file of numbers charts with the rows numbered 1, 2, 3.
- The first row is treated as a header when no cell in it is a number.
- JSON: an array of objects (`label`/`value` if present, otherwise detected the same way as CSV columns), an object of name to number, or an array wrapped in one property like `{"data": [...]}`.
- Numbers written for display (`"1,200"`, `$40`) are not read as numbers. That is #1198.
- Up to 100 data points.

## The part that makes this bigger than one file

There are 20 localised copies under `apps/docs/<locale>/tools/files/chart-maker.md`. Per CLAUDE.md these go through `pnpm i18n:translate` with its `--export` / `--import` handoff, not a direct edit, and never with `--dry-run`, which writes stub garbage into the locale files.

The one-line catalog description in `packages/shared/src/constants.ts:1781` and the 21 `packages/shared/src/i18n/*.ts` locale files was already updated by the #1158 fix. That catalog is a different surface from these docs pages and is edited directly. Don't confuse the two.

Out of scope for #1158 because pulling 21 docs pages and the translation pipeline into a backend bugfix would have tripled the diff.

Issue #1200, opened 2026-09-19
test(chart-maker): a stale CUSTOM_ROUTE_TOOLS entry keeps chart-maker out of both fixture matrices


snapotter-hq:
`chart-maker` sits in `CUSTOM_ROUTE_TOOLS` in two places:

- `tests/integration/generated/format-matrix-generated.shared.ts:56`
- `tests/integration/generated/format-matrix-multimodal.test.ts:52`

Both sets are described as tools that do not use the standard factory endpoint. chart-maker does use it. It calls `createToolRoute` like any other factory tool, and it is not in `REGISTRY_EXEMPT` (`tests/integration/tool-route-drift.test.ts:18`), which is the set that actually tracks non-factory contracts. The other four entries (`barcode-generate`, `qr-generate`, `passport-photo`, `html-to-image`) are generators with no file input, so they belong there. chart-maker doesn't.

## What it costs

The exclusion means no fixture in `tests/fixtures/data/valid/` ever reaches the tool through the matrices. `tiny-a.csv` is the one that stings:

```csv
id,name
1,alpha
```

A leading numeric id with a text column after it is exactly the shape #1158's detection had to get right. Feeding it through the matrix would have shown the behaviour change for free: before #1158 that file returned 422, now it returns 200 charting `alpha = 1`. Nobody saw that, because the tool isn't in the matrix.

`tiny.json` (`[{"name":"Ada","age":36}]`) is the same story on the JSON side.

## Fix

Drop `chart-maker` from both sets and run the matrices. The invariant they assert is loose on purpose (200, 202, 400, 413, 415, 422 all pass; a 500 or an undecodable success is the failure), so this should be a small change.

Expect to deal with one thing: the matrices feed image fixtures at everything, and chart-maker's `acceptedInputs` is `[".csv", ".json"]`, so a PNG should come back 415 from the extension gate at `apps/api/src/routes/tool-factory.ts:392`. Confirm that rather than assuming it, since #1158's telemetry showed PNG runs hitting chart-maker and failing, and nobody has checked which layer rejected them.

Worth checking whether the other four entries are stale too while you're in there.

Found while fixing #1158.

Issue #1199, opened 2026-09-19
fix(chart-maker): a CSV with years as column names charts its header row as data


snapotter-hq:
Drop this on `/files/chart-maker`:

```csv
region,2023,2024,2025
EMEA,10,20,30
APAC,5,6,7
```

Expected: two bars, or a rejection saying the shape isn't supported.
Actual: three bars. The first is labelled `region` with a value of 2025, and because 2025 dwarfs 30 and 7, it flattens the real data against the y-scale into two invisible stubs.

## Where

`parseCsv` (`apps/api/src/routes/tools/chart-maker.ts:184`) decides the first row is a header when no cell in it holds a number:

```ts
const hasHeader = rows.length > 1 && rows[0].every((cell) => numericCell(cell) === null);
```

Year columns are numbers, so `hasHeader` is false and the header row is charted as data.

Pre-existing, not a regression. The rule this replaced (`Number.isNaN(Number(rows[0][1]))`) got the same file wrong the same way, and #1158 called it out as a gotcha for whoever rewrote the detection. It survived the rewrite.

## Why it's hard

The obvious fix, comparing row 0's shape against the rows below, doesn't work here. Row 0 is `[text, num, num, num]` and so is every data row, so a pattern comparison sees no difference. That's also what makes this file genuinely ambiguous: `region,2023,2024,2025` is wide format where the columns are the series, and charting it properly means transposing, not picking two columns.

Two routes:

1. Detect wide format and transpose. A first row of a text cell followed by all-numeric cells, where the numbers look like years or dates, is a decent signal. Then each data row becomes a series.
2. Don't chart it, and say why. "This looks like a pivot table with years across the top. Chart Maker plots one label column against one number column." Cheaper, and honest.

Route 2 is probably right first. Route 1 changes what the tool is.

## Gotchas

- Don't tighten `hasHeader` to "row 0 has at least one text cell", because `region` is a text cell and the row still isn't recognised.
- Header detection has to stay one pass. The earlier two-pass idea (detect columns, then re-check row 0 against the chosen value column) is circular, since the header row participates in the column vote that picks the column you'd check it against.
- `tests/integration/tools/data/chart-maker.test.ts` has a `Chart Maker column detection` block (from `:283`) using PNG-byte equality: render the awkward file and its hand-shaped equivalent, compare bytes. That technique pins which columns were picked and is the right tool for this one.

Found while fixing #1158.

Issue #1198, opened 2026-09-19
fix(chart-maker): numbers written the way a spreadsheet shows them are dropped or charted from the wrong column


snapotter-hq:
chart-maker reads a cell as a number with `Number()`, so anything a spreadsheet formats for display is not a number to it. `"1,200"`, `$40`, `85%`, `(500)` all come back `NaN`.

Two ways that bites, both landing on a 200.

## Rows vanish

`tableToPoints` (`apps/api/src/routes/tools/chart-maker.ts:113`, the skip is at `:174`) drops a row whose value cell holds no number. That is right for blanks. It is wrong for a thousands separator, which by definition only shows up on values over 1000, so the rows that disappear are the biggest ones.

```csv
month,sales
Jan,900
Feb,"1,200"
Mar,"2,400"
Apr,850
May,700
Jun,650
```

Expected: six bars, Mar tallest.
Actual: four bars. Feb and Mar are gone, the chart shows the four worst months, and it looks like a perfectly normal four-bar chart. Nothing tells the user two rows were dropped.

The value column still wins its vote because the vote needs a strict majority, and 4 of 6 clears it. Shrink the file to four rows with two formatted values and the column drops to 2 of 4, misses the majority, and the whole file is rejected instead. Either way the user is misled.

## The wrong column gets charted

When the intended measure is unparseable, detection falls through to whatever else is numeric.

```csv
id,region,revenue
1,EMEA,$4400
2,APAC,$3610
```

Expected: a rejection, or revenue charted.
Actual: `id` is the only numeric column, so the chart is EMEA=1, APAC=2. Meaningless, and returned as a success. Before the #1158 fix this was a clean 422, because the value was hardcoded to column 2 and `EMEA` is not a number.

## Fix

Normalise in `numericCell` before `Number()`: strip a leading currency symbol, strip thousands separators, read a trailing `%`, read `(500)` as -500.

The trap is locale. `4.400,00` is four thousand four hundred in de-DE and four-point-four in en-US, and the file carries no locale, so guessing cell by cell will get some files wrong in a new way. Safer shape: decide the separator convention once per column by looking at every cell in it, then apply that one reading to the whole column. A column of `1,200` and `2,400` is unambiguous once you see them together.

Worth doing either way: when rows are dropped, say so. There is no channel for that today. A skipped count on the job result would let the UI say "charted 4 of 6 rows".

## Gotchas

[...]

PR #1202, opened 2026-09-19
fix(chart-maker): detect the label and value columns instead of hardcoding them

### LINKED ISSUE #1158: fix(chart-maker): CSV label and value are hardcoded to columns 1 and 2, so 97% of real runs fail
chart-maker is the most broken tool in the product, and nobody has reported it.

90 days of `tool_used` on 2.2.0, QA-sweep instances excluded: **67 failures in 69 runs, 97.1%, and every one of the 17 instances that tried it hit a failure.** All classified `error_kind: input`, so the tool is refusing the file rather than crashing. By input format: CSV fails 61% (85 runs, 15 instances), JSON 50%, PNG 100%.

The sweep instances are worth noting because they mask this. With them included the rate reads 65.7%. They passed a hand-shaped fixture; real users don't have one.

## What happens

Drop any ordinary CSV on `/files/chart-maker`, for example a spreadsheet export:

```csv
date,region,units,revenue
2026-01,EMEA,120,4400
2026-02,EMEA,98,3610
```

Expected: a chart, or at minimum a prompt asking which column is the label and which is the value.
Actual: the job fails. Nothing in the UI ever said what shape was required.

## Why

`parseInput` in `apps/api/src/routes/tools/chart-maker.ts:82-85` hardcodes the two columns:

```ts
return rows.slice(start).map((row) => ({
  label: String(row[0] ?? ""),
  value: Number(row[1]),
}));
```

Column 0 is the label, column 1 is the value, always. The CSV above puts `region` in column 1, so every `value` is `NaN`, and the render fails downstream. A four-column CSV where the number happens to sit in column 3 fails the same way. There is no column picker in `apps/web/src/components/tools/chart-maker-settings.tsx`, and the catalog description at `packages/shared/src/constants.ts:1781` promises "Create bar, line, or pie charts from CSV or JSON data" with no shape caveat.

JSON has the same problem in the other direction (`chart-maker.ts:48-60`): it accepts an array of `{label, value}` or a flat object of key to number, and rejects everything else, including the very common `{"data": [...]}` wrapper.

## Fix

The honest fix is a column picker, and it is not large. Parse the file, send the header row and a few sample rows back to the client, and let the user map label and value. `find-duplicates` and `image-to-base64` already use `custom-results` to render tool-specific state, so there is a precedent for a two-phase tool in this codebase.

The cheap fix, if the picker is too much for now: auto-detect instead of hardcoding. Pick the first column whose values parse as numbers across the majority of rows as the value column, and the first non-numeric column as the label. That turns the example above into a working chart with no UI change, and it is maybe twenty lines in `parseInput`.

Either way, fix the failure message. `ToolInputError` with "chart-maker needs a label column and a numeric value column; column 2 of your file is 'region'" beats a generic processing failure.

Whichever route, narrow the catalog description so it stops over-promising.

## Gotchas

- The header-skip heuristic at `chart-maker.ts:77-79` (`Number.isNaN(Number(rows[0][1]))`) is doing double duty as header detection. Auto-detection has to replace it, not sit next to it, or a numeric header row will be treated as data.
- `acceptedInputs` is `[".csv", ".json"]` but `parseInput` branches on the filename suffix at `chart-maker.ts:44`. A `.json` file with CSV content takes the JSON path and throws a parse error. Low priority, but note it if you touch the dispatch.
- The PNG runs in the data (7 runs, all failed) mean something is getting a `.png` past `acceptedInputs`. Worth a glance, separate problem.

## Acceptance

- A four-column CSV with the numeric column anywhere produces a chart.
- A CSV with no numeric column at all fails with a message naming what was missing.
- `{"data": [{"label": ..., "value": ...}]}` either works or explains itself.
- The catalog description matches what the tool actually accepts.

snapotter-hq:
Closes #1158

`parseInput` read the label from column 1 and the value from column 2, always. Any CSV whose numbers sat anywhere else gave `NaN` for every row and a generic rejection: 67 failures in 69 runs over 90 days, every instance that tried the tool.

## What changed

The value column is the rightmost one holding numbers in more than half the rows. The label is the leftmost one that isn't. If every column is numeric, the first is the label and the last is the value; a single-column file numbers its rows 1, 2, 3.

**Rightmost, not the leftmost the issue sketched.** Leftmost breaks `id,name,sales` (charts 1, 2, 3), `year,sales` (charts years and labels them by sales), and `timestamp,value`, which is the shape every database and admin export produces. Rightmost breaks `product,revenue,year`, where trailing metadata follows the measure. Both fail somewhere, but leftmost fails by charting a key column that is monotonically increasing, which is useless on every row, while rightmost only fails on files that put metadata after their measure. The tradeoff is written into the comment above `tableToPoints` so the next person flipping the scan direction sees it.

Header detection moved with the columns: the first row is a header when nothing in it is a number. The old probe asked only whether column 2 was numeric, which is the same wrong question the bug came from.

JSON had the defect from the other side. It now unwraps a single array property (`{"data": [...]}`), detects keys across the union of every object so records omitting an optional field don't vanish, and still honours an explicit `label`/`value` pair when the file names one. Dispatch sniffs the content instead of the filename, so a `.json` file holding CSV stops dying in the JSON parser.

Four more, all reachable from the same function and all found by the review agents:

- `skipEmptyLines: "greedy"`. Plain `true` keeps rows of bare commas as `["",""]`, which every spreadsheet export trails and which dragged the numeric vote under a majority. `month,sales` with three trailing `,` rows was rejected with "none of these have numbers: month, sales", about a column that plainly had numbers.
- `UndetectableDelimiter` is forgiven only when the parse really did yield one column. Papa raises that code both for a genuine single-column file and when detection failed and the comma fallback split the data, so `sales / 1,200 / 2,400` was about to chart 200 and 400 instead of erroring.
- The no-numeric-column message is capped at 6 names and 260 characters. `friendlyError` (`apps/api/src/lib/errors.ts:66`) replaces anything over 280 with a generic sentence, so a 12-column file would have thrown away the column names that are the whole point of the message.
- `parseInput` failures that aren't `ToolInputError` are rethrown unchanged. The old blanket rewrap told the user their file was at fault, and `worker.ts:620` skips `logger.error` for input errors, so a bug in this code would never have reached Sentry.

The catalog description gained the 100-row cap in all 21 locales. `tool-dropzone.tsx` renders it on the drop target, so it lands where someone is about to drop a file.

I did not build the column picker the issue offers as the honest fix. That was the explicit call: a two-phase tool needs a new response shape, a `custom-results` display mode, strings in 21 locales, e2e coverage, and a sane default anyway for pipeline and batch runs where nobody is there to pick.

## Verification

Baseline on `ddad5956` before any change, `VITEST_MAX_FORKS=3`:

- `pnpm lint` 0, `pnpm typecheck` 0
- `pnpm test`: 2 failed, 886 passed. `conversion-presets > excel-to-csv` (90s test timeout plus a cascading hook timeout) and `ocr-runtime-dispatcher > terminates a request that exceeds its timeout` (fake-timer race). Both timing flakes, neither near this code.

After: `pnpm lint` 0, `pnpm typecheck` 0, `pnpm test` 1 failed, 887 passed. Not one of the baseline two: both of those passed this time, and the one that failed is `scim > concurrent PUT renames onto one displayName`, the known race flake from #980/#990/#1000/#1156.

Disjoint failure sets across two runs of the same tree is itself the flake signature, so I re-ran all three in isolation: SCIM 88 passed, conversion-presets and ocr-runtime-dispatcher 130 passed. Nothing here touches auth, SCIM, or Postgres locking. Zero failures attributable to this change.

A first full-fork `pnpm test` crashed the Docker VM and reported 799 failed suites on `ECONNREFUSED` from dead testcontainers. That run is discarded, not a baseline. Everything above is `VITEST_MAX_FORKS=3`.

**The tests have teeth.** Every one of the 18 new tests was run with the fix reverted. Nine fail against `main`. Seven more fail against the pre-review version of this branch. The two that pass on both ("still rejects a file whose only numeric column is negative", "prefers an explicit value key over a later numeric one") are preservation guards for behaviour deliberately kept, not proofs of a fix.

They assert *which* columns were picked, not just a 200: render the awkward file, render the hand-shaped equivalent, compare PNG bytes. Both buffers come from the same sharp build in the same process, so there is no stored baseline and no platform suffix to maintain.

**Callers.** `registerChartMaker` has exactly one, the registration array in `routes/tools/index.ts` (LSP find-refs). Everything else is module-private. No test asserts the chart-maker description text and there are no snapshots. Ran the registry, catalog and i18n suites (`tool-route-drift`, `catalog-integrity`, `i18n-parity`, `landing/tool-search`, `tool-registry`, `tool-registry-expanded`, `generated-fixtures`, `conversion-presets-kills`, `modality-helpers`): 160 passed. Ran `tests/integration/tools/data` plus `settings-matrix`: 639 passed, 58 skipped.

Three review agents ran over the diff. Working tree hash-checked before and after each; none of them touched it.

## Known limits, documented not fixed

- Numbers written the way a spreadsheet displays them (`"1,200"`, `$40`) still aren't numbers, so rows carrying them drop out or detection falls through to a worse column. #1198, and the highest-value follow-up.
- A CSV with years as column names charts its header row. Pre-existing, called out as a gotcha in #1158, and it survived the rewrite because row 0 and the data rows have the same numeric shape. #1199.
- chart-maker is excluded from both fixture matrices by a stale `CUSTOM_ROUTE_TOOLS` entry, which is why `tiny-a.csv` (`id,name` / `1,alpha`) never reached it. #1200.
- The docs page still describes the two-column contract this removes, with 20 localised copies behind the `i18n:translate` handoff. #1201.
- The 100-point cap is still applied after a full parse, and detection adds an `O(rows x cols)` pass. Measured at 212ms and 125MB for a 200k-row file, against a 100MB upload limit. `preview: 102` on the Papa call would bound it, at the cost of a file whose first 100 rows are junk. No user has reported it, so it stays a note here rather than an issue.


snapotter-hq:
Ran silent-failure-hunter, pr-test-analyzer and code-reviewer over the diff. All three replayed the parse functions against papaparse 5.5.4 rather than reasoning about them, so the inputs below are observed. Between them they found a dozen real problems in the first version of this branch, most of them things the fix itself introduced. What follows is every finding and where it went.

## Fixed here, after the review

Each of these has a test in the `Chart Maker column detection` block, and each was confirmed to fail against the pre-review commit.

- **`skipEmptyLines: true` rejected ordinary spreadsheet exports.** It only drops rows that are exactly `[""]`, so a line of bare commas survives as `["",""]`. `month,sales` with three trailing `,` rows scored 2/5 on the numeric vote, missed the majority, and got "none of these have numbers: month, sales" about a column that plainly had numbers. That is the #1158 failure mode surviving the #1158 fix. Now `"greedy"`.
- **The `UndetectableDelimiter` filter was about to chart wrong numbers.** Papa raises that code for a genuine one-column file and also when detection failed and the comma fallback split the data. `sales / 1,200 / 2,400 / 3,600` parses to widths `[1,2,2,2]` and would have charted 200, 400, 600. Forgiven now only when every row really did yield one field.
- **The new error message was being thrown away.** `friendlyError` (`apps/api/src/lib/errors.ts:66`) swaps anything over 280 characters for a generic sentence. Listing 12 column names blows that, so the wide files that most need the column names got the least useful message. Capped at 6 names and 260 characters, with a test asserting the length.
- **Our own bugs were being blamed on the user's file.** The catch around `parseInput` rewrapped every error as `ToolInputError`, and `apps/api/src/jobs/worker.ts:620` deliberately skips `logger.error` for input errors, so a `TypeError` in these 200 new lines would never have reached Sentry. Non-`ToolInputError` now rethrows unchanged.
- **Rightmost-numeric regressed two shapes that worked before.** `region,sales,growth` picked `growth`, hit the pre-existing negative-value guard, and 422'd a file that used to chart fine. Columns holding negatives are skipped now, keeping the last one as a fallback so an all-negative file still gets the specific message.
- **A stray extra field in one row of two hijacked the value column.** `name,value / A,1 / B,2,99` gave the phantom third column a share of exactly 1/2, which `>= 0.5` accepted, charting one bar of 99 and dropping row A. The threshold is a strict majority now, which is what the issue meant by "majority" anyway.
- **JSON key detection read only `objects[0]`.** `[{"name":"A"},{"name":"B","revenue":20}]` charted one point and silently dropped the rest. Unions the keys now.
- **A BOM'd `.json` file could never parse.** U+FEFF counts as `\s`, so the content sniff matched and `JSON.parse` then threw. Stripped.
- **`Object.keys({})` is truthy,** so `[{},{}]` produced "none of these have numbers: ." with nothing before the period.
- **Unbounded user data in an error message.** `JSON.stringify(item.value)` on a 1.1MB nested object produced a 1.1MB `Error.message` that went through three regex passes and into a log line. Sliced to 60.
- **A comment that lied.** It claimed blank-skipping handles the `Total` footer spreadsheets append. A `Total,300` row keeps its number and gets charted as an extra bar. Corrected, and it now points at #1198.
- **`JSON_SHAPES` promised a shape that was not implemented,** saying "either of those wrapped in a single property" when only arrays were unwrapped. Reworded to match the code.

## Considered and not done

- **Relaxing Papa's quote errors to non-fatal.** code-reviewer reported that `InvalidQuotes` and `MissingQuotes` come with usable data. Checked it: `product,sales / "Acme" 12-pack,4400 / Beta,1200` returns `[["product","sales"],["Acme\" 12-pack,4400\nBeta,1200\n"]]`, which is one mangled row, not full data. Keeping them fatal.
- **Defaulting the chart title to the value column's header.** Two reviewers asked for it, and they are right that a wrong column pick is invisible in a bare PNG. It is also a behaviour change for every chart that currently has no title, and #1158 asked for column detection, not retitling. Worth its own issue if someone wants it.
- **Accepting `{"data":[...], "errors":[]}`.** One predicate would do it. The acceptance criteria name `{"data": [...]}`, which works, and widening accepted input past that is not this PR.
- **Bounding the parse with `preview: 102`.** Measured at 212ms and 125MB for a 200k-row file, against a 100MB upload cap, and not a regression. CLAUDE.md puts speculative optimisations with no user report in the PR body, so it is there.
- **Exporting `parseInput` so the boundary cases become unit tests.** Fair structural point, and a refactor beyond this issue.
- **Header detection eating a first row whose value cell is blank.** Traced it: `Jan,` is dropped either way, because `""` is not numeric and the row loses the value-cell check regardless. Tightening the rule changes nothing.
- **Pinning the `product,revenue,year` tradeoff with a test.** That would enshrine a known-wrong output as intended behaviour. The comment above `tableToPoints` owns the tradeoff instead.
- **`1.5e308` summing to `Infinity` in `renderPieSvg`**, which renders a blank canvas. Exotic enough to leave alone; `numericCell` already blocks a literal `Infinity`.

## Filed

- #1198, numbers written the way a spreadsheet displays them. The highest-value one left.
- #1199, a CSV with years as column names charts its header row.
- #1200, a stale `CUSTOM_ROUTE_TOOLS` entry keeps chart-maker out of both fixture matrices, which is why `tiny-a.csv` never reached it.
- #1201, the docs page still describes the two-column contract this removes.

Working tree was hash-checked before and after each agent. None of them modified it.

Issue #1205, opened 2026-09-19
fix(ai): an uppercase DATA: URI is dropped from html-to-pdf output


snapotter-hq:
RFC 3986 says URI schemes are case-insensitive, so `DATA:image/png;base64,...` is a perfectly legal data URI. The WeasyPrint fetcher in `doc_html_pdf.py` matches with `url.startswith("data:")`, which is case-sensitive, so an uppercase or mixed-case spelling is refused and the image is dropped from the PDF.

Found while fixing #1157, and pinned by a test there rather than fixed. The behaviour predates that PR: the old inline fetcher used the same `startswith`. Fixing it loosens the input handling of the one control that keeps this path off the network, which deserves its own review rather than riding along in a bug fix.

## Repro

```python
import importlib.util
spec = importlib.util.spec_from_file_location("d", "packages/ai/python/doc_html_pdf.py")
m = importlib.util.module_from_spec(spec); spec.loader.exec_module(m)
fetch = m._make_url_fetcher(lambda url, *a, **k: {"string": b"served"})
fetch("DATA:image/png;base64,AA==")
# ValueError: remote resources are disabled: DATA:image/png;base64,AA==
```

Expected: served, the same as lowercase `data:`.

Actual: refused. WeasyPrint logs the refusal and lays the page out without the image, so the document still converts. The user gets a PDF quietly missing a picture that is embedded in their own file and never touched the network.

## Where it lives

`packages/ai/python/doc_html_pdf.py:28`, inside the closure `_make_url_fetcher` returns:

```python
if url.startswith("data:"):
```

## The fix, and why it is not a one-character change

`url[:5].lower() == "data:"` is the whole edit. What needs thought is the direction: this is the single network chokepoint for html-to-pdf, markdown-to-pdf and epub-convert's pdf format, and every change to it widens or narrows what gets served without a fetch. Case-folding is safe on its own reasoning (a `data:` URI carries its payload inline and reaches no host), but it should land with someone having deliberately confirmed that rather than as a drive-by.

Whoever takes it also needs to flip the assertion that currently pins the present behaviour, in `tests/unit/ai/ssrf-url-fetcher.test.ts`, where `"DATA:image/png;base64,AA=="` sits in the refuse list with a comment calling the fail-closed posture deliberate.

## Gotchas

- Keep it a prefix comparison. Lowercasing the whole URL to test it would be wasteful on a large base64 payload, and it must not be the string that gets passed on.
[...]

Issue #1204, opened 2026-09-19
fix(doc-engine): an empty error message from a doc_* script reads as success


snapotter-hq:
`str()` is the empty string for any exception built with no arguments, `MemoryError` among them. Every `doc_*` script prints `{"error": str(exc)}` on failure, and every wrapper in `packages/doc-engine/src/python-docs.ts` then tests that message for truthiness. An empty one is falsy, so the wrapper returns normally and the job is recorded as a success that produced nothing.

Found while fixing #1157. Fixed there for `doc_html_pdf.py` only, by emitting `str(exc) or type(exc).__name__`, because that script's exposure went up: documents that used to be rejected by a pre-scan before WeasyPrint ever parsed them now reach the renderer, and those are exactly the heavy ones. The other nine wrappers and the falsy check itself were out of scope.

## Repro

```python
>>> str(MemoryError())
''
```

Then, with a `doc_*` script raising a bare `MemoryError`: `main()` prints `{"error": ""}` and exits 1, the dispatcher hands the stdout back, and `parseDocsJson` returns `{error: ""}`.

Expected: the job fails with a message saying the document was too large.

Actual: the wrapper's `if (parsed.error)` is false, so it returns as though the conversion worked. The route then hands back a `scratchPath` the script never finished writing, and the user sees an ENOENT or an artifact-size mismatch from the worker instead of the real cause.

## Where it lives

- `packages/doc-engine/src/python-docs.ts` has 20 of these truthiness checks. `doc_flatten` at line 78, `doc_scrub_meta` at 91, `doc_redact` at 114, `doc_text` at 133, `doc_to_word` at 152, `doc_html_pdf` at 203, and so on. `doc_pagecount` at line 68 is the only one that happens to be safe, because it also requires `typeof parsed.pages === "number"`.
- `packages/ai/src/bridge.ts:1022`, `runDocsScript`, returns stdout and stderr and drops the exit code, even though the dispatcher does send one (`packages/ai/python/dispatcher.py:410`) and the non-dispatcher path checks it at `packages/ai/src/bridge.ts:466`. So the wrapper has no second signal to fall back on.

## Two ways to fix it, probably both

1. Check the key, not the value: `if ("error" in parsed)`. One-line change per wrapper, catches every producer including ones not yet written.
2. Have `runDocsScript` surface the dispatcher's exit code so a non-zero exit fails regardless of what landed on stdout. Bigger, and it makes the whole `doc_*` family fail closed.

[...]

Issue #1203, opened 2026-09-19
fix(ai): a refused remote resource leaves no trace, so a PDF with holes reports success


snapotter-hq:
A remote image or stylesheet in an HTML, Markdown or EPUB conversion is now dropped without a trace. The PDF has a hole where the picture was, the job row says completed, the API log is clean. Nobody, user or operator, can tell "this book had no cover" from "this book's cover lives on a CDN".

Found while fixing #1157, which removed the pre-scan that used to abort these conversions outright. Not fixed there: #1157 was about documents failing to convert at all, and no warning channel exists anywhere on this path to put the news in.

## Repro

Convert an HTML file holding one remote image:

```html
<!doctype html><html><body><p>before</p>
<img src="https://example.com/tracker.png" alt="pic"><p>after</p>
</body></html>
```

Expected: the conversion succeeds, which is correct, and something somewhere records that a resource was refused.

Actual: `{"ok": true}`, a PDF with the alt text standing in for the image, no log line, nothing on the jobs row.

## Why nothing is logged

WeasyPrint does log it, at ERROR level:

```
Failed to load image at 'https://example.com/tracker.png': ValueError: remote resources are disabled: ...
```

But `weasyprint/logger.py` line 19 attaches a `logging.NullHandler()` to the `weasyprint` logger, which is the standard library-author convention, and the docs dispatcher configures no logging of its own. Verified on weasyprint 69.0: a render with a blocked `<link rel="stylesheet">` and a blocked `<img>` wrote zero bytes to stderr.

## Where it lives

- `packages/ai/python/doc_html_pdf.py:30`, inside the fetcher `_make_url_fetcher` builds. It raises `ValueError` and tells the caller nothing; WeasyPrint swallows it by design.
- `packages/ai/python/doc_html_pdf.py:66`, the `print(json.dumps({"ok": True}))` contract. One boolean, no room for a notice.
- `packages/doc-engine/src/python-docs.ts:192`, `htmlToPdfPy`, returns `void`.
- `apps/api/src/routes/tools/html-to-pdf.ts:28`, `apps/api/src/routes/tools/markdown-to-pdf.ts:28` and `apps/api/src/routes/tools/epub-convert.ts:47` all return `{scratchPath, filename, contentType}`, and `ToolJobResult` has no warnings field for a route to fill.

## Prior art worth cribbing

[...]

Issue #1174, opened 2026-09-16
test: the weasyprint-gated integration specs are skipped in every automated environment


snapotter-hq:
Found while fixing #983. Not fixed there: that PR only corrects one assertion, and this needs weasyprint in a test toolchain.

Three integration blocks sit behind `describe.skipIf(!pythonWith("weasyprint"))`:

- `tests/integration/tools/document/html-to-pdf.test.ts`, the whole describe (line 45)
- `tests/integration/tools/document/markdown-to-pdf.test.ts` (also needs the `markdown` module)
- `tests/integration/tools/document/epub-convert.test.ts`, the "epub-convert pdf" block (line 167, also needs pandoc)

`pythonWith()` in `tests/helpers/python-gate.ts` spawns the resolved interpreter (`.venv/bin/python3` if the repo has one, else `python3` on PATH) with `-c "import weasyprint"`. Nothing installs that module anywhere the suite runs:

- CI integration shards (`.github/workflows/ci.yml`): the shards fetch pandoc (line 215) but no Python packages. `setup-python` only appears in the pip-audit job.
- `docker/Dockerfile.test`: no pip install at all. `pnpm test:docker` runs the full vitest suite in that image, so the gate stays shut there too.
- Dev Macs: nobody has it, per CLAUDE.md and the header of html-to-pdf.test.ts.

There's no e2e fallback either. `grep -rl "html-to-pdf" tests/e2e tests/e2e-docker` comes back empty, so the header comment in html-to-pdf.test.ts ("the Docker compose smoke is the real end-to-end proof") describes coverage that doesn't exist.

Expected: the html-to-pdf and markdown-to-pdf SSRF contracts and the epub-to-PDF chain run on at least one automated lane.

Actual: they're skipped on every PR, every nightly and every docker run. #983 is the proof. `expect(row?.error).toMatch(...)` against a jsonb object sat in the html-to-pdf SSRF spec from #649 (2026-07-27) until 2026-09-16 and nothing ever went red. The regex prescan has unit coverage in `tests/unit/ai/ssrf-prescan-regex.test.ts`; the route-to-worker-to-jobs-row contract only lives in these skipped blocks.

[...]

PR #1206, opened 2026-09-19
fix(ai): let WeasyPrint omit remote resources instead of rejecting the document

### LINKED ISSUE #1157: fix(doc-engine): remote-ref pre-scan rejects documents that only link to a URL, breaking epub-convert, markdown-to-pdf and html-to-pdf
Three tools share `doc_html_pdf.py`, and all three sit near the top of the fleet's failure table. 90 days of `tool_used` on 2.2.0, with the v2.2.0 QA sweep instances excluded:

| tool | runs | fail % | instances hit | dominant error |
| --- | --- | --- | --- | --- |
| epub-convert | 63 | 69.8% | 9 of 17 | `exit-1` (39) |
| markdown-to-pdf | 52 | 38.5% | 4 of 22 | `exit-1` (18) |
| html-to-pdf | 204 | 22.5% | 10 of 16 (62%) | `exit-1` (42) |

They fail for one reason: the pre-scan in `packages/ai/python/doc_html_pdf.py` treats a hyperlink as a remote resource.

## Repro

```python
import importlib.util
spec = importlib.util.spec_from_file_location("d", "packages/ai/python/doc_html_pdf.py")
m = importlib.util.module_from_spec(spec); spec.loader.exec_module(m)
m._find_remote_refs('<p>See <a href="https://example.com">docs</a></p>')
# ['https://example.com']  -> sys.exit(1), whole conversion aborts
```

Expected: the PDF renders with the link intact as text, and nothing is fetched.
Actual: `exit-1`, and the user sees a processing failure with no way around it.

In product terms, a markdown file containing a single `[docs](https://example.com)` cannot be converted to PDF. Neither can an HTML page with one external anchor, nor an EPUB with a link in its colophon, which is most EPUBs.

## Why it happens

`_REMOTE_REF_RE` at `packages/ai/python/doc_html_pdf.py:12` matches six attributes:

```
src | href | action | srcset | poster | formaction
```

Only `src`, `srcset` and `poster` cause WeasyPrint to fetch anything. `href` on an `<a>` is navigation, `action` and `formaction` are form submission targets, and none of the three is dereferenced during render. They are matched anyway, and `doc_html_pdf.py:67-70` turns any hit into `sys.exit(1)` for the whole document.

Markdown is the worst case because the tool builds the HTML itself at `doc_html_pdf.py:61`: `md.markdown()` rewrites `[text](https://...)` into `<a href="https://...">`, so the scan rejects a document the user never wrote any HTML for.

The scan is also redundant for the thing it protects. The module's own docstring says so:

> The url_fetcher remains as a defense-in-depth backstop: it blocks any reference that slips past the scan, raising ValueError so WeasyPrint omits the resource with zero outbound requests.

`no_remote_fetcher` at `doc_html_pdf.py:46` is the real SSRF control, and it is unconditional. The pre-scan adds no safety over it. What it adds is converting "omit one remote image" into "fail the entire document".

## Fix

In order:

1. Drop `href`, `action` and `formaction` from `_REMOTE_REF_RE`. They cannot produce an outbound request, so rejecting on them is a pure false positive. This alone should clear most of the epub and markdown failures.
2. Decide whether a genuinely fetchable remote ref (`src`, `srcset`, `poster`, `url()`, `@import`) should still abort. My lean is no: let `no_remote_fetcher` do its job and render the document with the remote image missing, the way a browser with images disabled behaves. A book that fails to convert because of one tracking pixel is worse than a book that converts without it. If we keep the hard fail, it needs to be a `ToolInputError` naming the offending URL, not a bare `exit-1` that surfaces as "processing failed".

Verify the fetcher actually omits rather than aborting the render before committing to option 2. WeasyPrint's behaviour when a `url_fetcher` raises has changed across versions, and the docstring's claim is untested. If a raise does kill the render, return a 1x1 transparent PNG from `no_remote_fetcher` instead of raising.

## Gotchas

- Keep the SSRF posture intact. The point of #706 and #707 was that nothing in this path reaches the network. Relaxing the scan must not relax the fetcher.
- `data:` URIs already pass through `default_url_fetcher` at `doc_html_pdf.py:47`. Don't break that.
- `epub-convert` reaches WeasyPrint through pandoc (`--standalone`, no `--embed-resources`, see `apps/api/src/routes/tools/epub-convert.ts:42`). The HTML pandoc emits is what gets scanned, not the EPUB, so a fixture needs to survive that hop.

## Acceptance

- A markdown file with an inline external link converts to PDF, link rendered as text.
- An HTML page with `<a href="https://...">` converts.
- An EPUB whose spine contains external links converts.
- A document with `` either converts with the image absent, or fails with a message naming the URL. Not `exit-1`.
- No outbound request in any of the above. Assert on the fetcher, not just the exit code.

### LINKED ISSUE #1175: test: format-matrix-multimodal feeds remote-img.html to html-to-pdf as a clean-status case and fails once weasyprint is present
Found while fixing #983, on a Mac with weasyprint installed in a repo-root `.venv`. Not fixed there: #983 is a one-assertion change in a different spec, and this needs a decision about how the matrix treats edge fixtures.

Repro (needs weasyprint importable by the python3 that `tests/helpers/python-gate.ts` resolves):

```
pnpm vitest run tests/integration/generated/format-matrix-multimodal.test.ts -t "html-to-pdf"
```

Expected: every html-to-pdf case in the matrix ends in a clean success or a clean rejection, and the accounting case passes.

Actual, two failures:

```
FAIL  multi-modality tool x format matrix > html-to-pdf > remote-img.html -> clean status
Error: remote resources are disabled: http://127.0.0.1:9/x.png

FAIL  multi-modality tool x format matrix > html-to-pdf > conserves generated case accounting
Error: html-to-pdf: generated accounting is not conserved (attempted=3, accepted=2, rejected=0, skipped=0)
```

Where it lives:

- `tests/helpers/generated-fixtures.ts` line 28: `generatedFixtureDirectories()` includes `fixtureDir.document.edge`, so `tests/fixtures/document/edge/remote-img.html` is fed to html-to-pdf as a should-succeed input.
- `tests/integration/generated/format-matrix-multimodal.test.ts` around line 320: the tool returns 202, `waitForGeneratedJobArtifact()` then throws because the job failed, and `isEngineUnavailableFailure()` in `tests/helpers/generated-case-accounting.ts` only recognises missing ffmpeg/ffprobe/pdfcpu/soffice/pandoc binaries. Nothing calls `accounting.reject()` for a job that was accepted and then legitimately failed, so the case throws and the accounting is off by one.
- The fixture exists to be rejected: it's the input for the SSRF contract in `tests/integration/tools/document/html-to-pdf.test.ts`, and the prescan in `packages/ai/python/doc_html_pdf.py` (line 69) rejects it by design.

Nobody sees this today because weasyprint is absent from dev Macs, the CI shards and `docker/Dockerfile.test`, so `findMissingGeneratedPythonPrerequisite()` skips the whole html-to-pdf block (#1174). Whoever opens that gate hits this first.

Gotchas:

- The matrix contract is "no 500s: only clean success or clean rejection (200, 202, 400, 413, 415, 422, 501)". A 202 whose job then fails with a tool-level rejection has no bucket. Any fix has to pick one, and `GeneratedCaseAccounting` in `tests/helpers/generated-case-accounting.ts` has to agree with it.
- `document/edge/` also holds other fixtures; check which tools they reach before excluding the directory wholesale.
- #1157 (the prescan rejects any `href`) widens what counts as "rejected by design".

Starting sketch: teach the async branch to classify an expected tool-level failure. Either count a failed job as `accounting.reject()` when the persisted error is a `ToolInputError`-class message, or keep `edge/` out of the clean-status matrix and give edge fixtures their own expected-outcome table. The first keeps the coverage; the second is smaller.

snapotter-hq:
A Markdown note containing `[docs](https://example.com)` could not be turned into a PDF. Neither could an HTML page with one external anchor, nor most EPUBs, since a colophon link is close to universal. Over 90 days on 2.2.0 that took out 70% of epub-convert runs, 38% of markdown-to-pdf and 22% of html-to-pdf.

## What broke

The pre-scan in `doc_html_pdf.py` matched six attributes, three of which no renderer dereferences: `href` on an `<a>`, `action` on a `<form>`, `formaction` on a `<button>`. Any hit aborted the whole document with `sys.exit(1)`, which reaches the user as a bare processing failure. Markdown was the worst case, because `md.markdown()` writes the `<a href>` itself from a link the user typed as Markdown.

## Why the scan went rather than getting narrower

Dropping those three attributes was the obvious fix and it is not enough. Checking what weasyprint 69.0 actually reads showed the scan was wrong in both directions.

| ref | pre-scan | WeasyPrint fetches |
| --- | --- | --- |
| `<a href>`, `<form action>`, `<button formaction>` | rejected | no |
| `<link rel=stylesheet href>` | rejected | **yes** |
| `<svg><image href>` | rejected | **yes** |
| ``, `srcset`, `poster`, `url()`, `@import` | rejected | yes |
| `@font-face src: url()` | **missed** | yes |
| `<object data>` | **missed** | yes |
| `file:///etc/passwd` | **missed** | yes |

So `href` cannot simply be dropped, because two fetchable refs use it. Telling `<a href>` from `<link href>` means parsing HTML with a regex, which is the bug this issue is about. Meanwhile the fetcher already saw every row in that table, including the three the scan never covered, and it was already unconditional. The scan's only effect was turning "omit one tracking pixel" into "this book cannot be converted".

## Verification the ticket asked for

WeasyPrint's behaviour on a raising `url_fetcher` has moved across releases and the docstring's claim was untested, so the issue made removing the hard fail conditional on checking it. With `socket.create_connection` and `socket.socket.connect` patched to record and raise, every case rendered and none opened a socket: remote ``, remote `<link>`, `@import`, `url()` background, `@font-face`, `<svg><image>`, `<object data>`, `file://`. The guard itself was proved live by handing the same document to `default_url_fetcher`, which recorded a connection to 127.0.0.1:9.

End to end through the real pandoc hop, the epub fixture's intermediate HTML converts to a 7 KB PDF whose text reads "See the publisher." where main exits 1. That last check was `pdftotext` by hand, not an assertion: WeasyPrint compresses the content stream, so there is no cheap way to assert on rendered text without adding another binary gate to a suite that already has two.

## Two things found while doing it

**A fetcher that imports is a fetcher that fails silently.** Moving `_no_remote_fetcher` to module level with a lazy `from weasyprint.urls import default_url_fetcher` inside its `data:` branch looked harmless. It is not: WeasyPrint converts any exception a fetcher raises into an omitted resource, so an ImportError there is swallowed once per resource. Deleting the symbol gives `{"ok": true}`, exit 0, empty stderr, and an invoice PDF with a hole where the embedded logo was. Not hypothetical either, since `default_url_fetcher` carries a deprecation warning naming 69.0, which is what the Dockerfile pins. `main()` now imports it inside the existing guard and `_make_url_fetcher` closes over it, so a missing symbol stops the job instead.

**An empty error message reads as success.** `str()` is `""` for any exception built with no args, `MemoryError` included, and `htmlToPdfPy` tests that message for truthiness. The error branch now emits `str(exc) or type(exc).__name__`. That path is newly load-bearing: heavy documents used to be rejected before WeasyPrint ever parsed them. The truthiness check itself, and the other nine `doc_*` wrappers, are #1204.

## Tests

`tests/unit/ai/ssrf-prescan-regex.test.ts` tested regexes that no longer exist. `tests/unit/ai/ssrf-url-fetcher.test.ts` replaces it and pins the contract that does. It stubs weasyprint and markdown in `sys.modules`, so unlike the gated integration blocks it runs in CI: the fetcher refuses `https`, `http`, the single-slash `http:/` pandoc emits, `file://`, `ftp://`, a protocol-relative ref and an uppercase `DATA:`; it serves `data:`; `main()` installs it in both modes and leaves every ref shape the pre-scan used to reject in the HTML; the module imports on an interpreter where weasyprint is unusable; and a refusal escaping `write_pdf` names the URL rather than exiting bare.

Three weasyprint-gated integration tests moved from asserting rejection to asserting conversion. Both SSRF contracts now point the document at a live recorder (`tests/helpers/request-recorder.ts`) and assert zero requests and zero connections after a mandatory reachability probe. A closed port only proves the fetch failed, which is also what a missing guard produces; the epub one was green with the guard deleted entirely before this change.

`.gitignore` gets `.venv`, since two test headers now tell contributors to create one and it was not ignored. The Stryker scope guard requires anything gitignored to be in `ignorePatterns`, so both root configs got it too.

## Verification

Run locally with the gate open, `.venv` symlinked to a venv holding weasyprint 69.0 and markdown 3.10.2, the Dockerfile pins. The suite ran in four chunks rather than one `pnpm test`, because a full run would not fit in memory alongside the Docker VM.

| chunk | baseline | after |
| --- | --- | --- |
| `pnpm lint` / `pnpm typecheck` | exit 0 | exit 0 |
| unit + packages | 10256 passed, 0 failed | 10270 passed, 0 failed |
| integration/tools/document | 162 passed, 0 failed | 164 passed, 0 failed |
| integration/tools (image, video, audio, data) | 2456 passed, 0 failed | 2456 passed, 0 failed |
| integration/platform, security, fixtures | 2590 passed, 1 failed | 2590 passed, 1 failed |
| integration/generated | 4968 passed, 2 failed | 4970 passed, 0 failed |

Zero new failures. The one in the platform chunk is the SCIM concurrency flake from #980 / #990 / #1000 / #1156, which hit a different test in the same suite each run and touches nothing this PR changes. The two that cleared are #1175: `remote-img.html` was fed to html-to-pdf as a should-succeed input and then rejected, which also threw the accounting off by one. It now ends in a clean success.

Teeth: reverting only `doc_html_pdf.py` turns the four new assertions red and leaves the rest green.

## What this does not cover

Between this merge and #1174, no automated lane runs the weasyprint-gated suites, so the end of the contract that says WeasyPrint really routes everything through the fetcher is verified locally and by reading its source, not by CI. The new unit suite runs in CI but stubs WeasyPrint, so it can only prove the fetcher refuses and that `main()` installs it.

A refused resource is also now invisible. `weasyprint/logger.py` attaches a `NullHandler`, so its ERROR line never reaches the dispatcher's stderr, and the script contract has no field for a notice. That is #1203, and it wants a warnings channel through the Python contract, `htmlToPdfPy` and `ToolJobResult` rather than a log line bolted on here.

One behaviour change worth naming: the output PDF now embeds the anchor as a clickable URI action. No server-side fetch, and it is the point of the issue, but such a PDF could not previously exist because the document was rejected outright.

Follow-ups: #1203, #1204, #1205.

Closes #1157
Closes #1175


snapotter-hq:
Ran silent-failure-hunter, code-reviewer and pr-test-analyzer over the diff before opening this. What they found, what changed, and what I dismissed.

## Fixed in this PR

**A fetcher that imports is a fetcher that fails silently.** The first draft had `_no_remote_fetcher` at module level doing a lazy `from weasyprint.urls import default_url_fetcher` inside its `data:` branch. WeasyPrint converts any exception a url_fetcher raises into an omitted resource, so an ImportError there is swallowed once per resource. Reproduced by deleting the symbol: `{"ok": true}`, exit 0, empty stderr, and an invoice PDF missing its embedded logo. `default_url_fetcher` also carries a deprecation warning naming 69.0, the version the Dockerfile pins, so this was on a timer. Now `main()` imports it inside the existing guard and `_make_url_fetcher` closes over it. Same probe gives exit 1 and `weasyprint not usable: cannot import name 'default_url_fetcher'`.

**The epub SSRF assertion was vacuous.** Its fixture pointed at a closed port, so the default fetcher and the blocking one produce identical outcomes. Deleting `url_fetcher=` from the script entirely left that test green. It now points at a live recorder and asserts zero requests and zero connections. This is the path that matters most, since it is the only test covering the pandoc hop.

**An empty error message reads as success.** `str()` is `""` for an exception built with no args, `MemoryError` included, and `htmlToPdfPy` tests it for truthiness. Fixed in this script's error branch, because its exposure went up: documents that used to be rejected before WeasyPrint parsed them now reach the renderer. The check itself and the other nine `doc_*` wrappers are #1204.

**The empty-request assertion had no positive control.** An unreachable listener would have read as a pass. The recorder now requires a `probe()` that proves reachability before any empty-log assertion, and counts sockets as well as completed requests so a connect-then-abort is visible.

**The markdown assertion was a tautology.** The stub ignored its input and returned a fixed anchor, which the assertion then looked for. It now echoes its input and the test asserts the real source reached `md.markdown`.

**Four `--` em-dash stand-ins** in the Python docstring and a test header, which CLAUDE.md bans outright. Gone.

**The lazy-import docstring gave a reason that was not true.** The test driver stubs `sys.modules`, so it would have been satisfied by a module-level import either way. Rewritten, and cut from twelve lines to seven.

**Duplication.** Rather than add a third hand-rolled `convertAndDownload`, both rewritten suites use the existing `waitForDownloadedJobArtifact` in `tests/integration/settle-job.ts`. That also brings size and MIME verification and cancel-on-timeout, none of which the hand-rolled poll did.

**Coverage the deleted regex suite had and the first replacement did not**: the CSS ref shapes (`@import`, `url()`, `@font-face`), plus `srcset`, `poster` and `<object data>`, are now in the unit fixture, so a pre-scan narrowed to CSS cannot sneak back. Added a no-stubs import check so the "imports without weasyprint" claim is tested, and a case for a refusal escaping `write_pdf` so the acceptance clause about never being a bare `exit-1` is pinned.

**`.venv` was not gitignored** while two test headers now tell you to create one. Added without a trailing slash, since it is normally a symlink. The Stryker scope guard then failed, correctly, because anything gitignored must be in `ignorePatterns`; both root configs got it.

## Considered and dismissed

**"Delete the orphaned `remote-img.html` fixture."** It is not orphaned. `generatedFixtureDirectories()` includes `tests/fixtures/document/edge/`, so the format matrix feeds it to html-to-pdf, and that is exactly the #1175 case this PR turns green. The two index aliases pointing at it do now have no direct consumer, because the SSRF tests need a dynamic port and build their HTML inline. `fixtures.security.htmlSsrf` was already unused before this diff, so removing one and leaving the other is worse than leaving both.

**"Assert the link renders as text in the PDF."** It is an acceptance criterion and I checked it by hand (`pdftotext` gives "See the publisher." and "See docs for more."), but there is no cheap automated version: WeasyPrint compresses the content stream, so the URL is not in the bytes, and extracting text needs fitz or poppler. That would add a third binary gate to a suite that already has two and runs nowhere automated.

**"Add egress assertions to markdown-to-pdf too."** All three tools share one chokepoint, proven in html-to-pdf and now epub-convert. A third copy asserts the same function again.

**"Guard `expect(hasPython).toBe(true)` under CI."** Fair, but it applies to every `skipIf(!hasPython)` suite in the repo, not this one, and it belongs with #1174.

**"Switch to `URLFetcher(allowed_protocols=("data",))`."** Cleaner API and a second scheme gate, but it does not exist below weasyprint 69, which would break native installs that are not on the exact Docker pin. Staying on `default_url_fetcher`; the loud-failure fix above handles its eventual removal.

**"Case-fold the `data:` check."** Real bug, filed as #1205 rather than fixed here. It loosens the input handling of the one control keeping this path off the network, which deserves its own review rather than riding along in a bug fix.

## Still open, and worth knowing before merging

Between this and #1174, no automated lane runs the weasyprint-gated suites. The CI-runnable unit suite stubs WeasyPrint, so it proves the fetcher refuses and that `main()` installs it, not that WeasyPrint routes everything through it. That end is verified locally and by reading weasyprint 69.0's source, where every dereference goes through `fetch(url_fetcher, url)` in `urls.py`.

PR #1207, opened 2026-09-19
fix(editor): close the polygonal lasso on Enter or a click on the first vertex

### LINKED ISSUE #1059: [Bug] Polygonal lasso only closes on double-click, with no hint saying so
Reported in discussion #828 by @CodecNomad: "Pressing enter doesn't do anything. Clicking on the initial point doesn't do anything either, how do you use it??"

The answer is double-click. Nothing in the UI says so, and the two gestures they tried are the two every other editor supports.

## Repro

1. Open `/editor`, load an image.
2. Press `L`, then `Shift+L` to reach Polygonal Lasso (or pick it from the options bar).
3. Click three or four times to drop vertices.
4. Press Enter. Nothing happens.
5. Click the first vertex again. Nothing happens; it just drops another vertex on top of it.

Expected: any of double-click, Enter, or clicking the start vertex closes the polygon.
Actual: only double-click does.

## Where it lives

- `apps/web/src/components/editor/tools/selection-tool.tsx:513-520` - `onMouseUp` deliberately does not close the polygon:
  ```
  // Polygonal lasso: mouseUp does NOT close the polygon, only dblclick does.
  ```
- `apps/web/src/components/editor/tools/selection-tool.tsx:559-568` - `onDoubleClick`, the only close path.
- `apps/web/src/hooks/use-editor-shortcuts.ts:701` - Enter is bound, but scoped to applying crop and transform. It never reaches the selection tool.
- `apps/web/src/components/editor/options/selection-options.tsx:177` - the options bar renders `t.editor.options.selection.polygonalLasso` as a bare label with no hint.

## Three parts, all in scope

1. **Enter closes the polygon** when the polygonal lasso is mid-draw. Needs to take priority over the crop/transform binding, or at least not be swallowed by it.
2. **Clicking the start vertex closes it.** Hit test against the first point with a tolerance in screen pixels, not canvas pixels, so it stays clickable when zoomed out. A visible target on the first vertex once there are enough points to close is the usual affordance.
3. **An on-screen hint** while the tool is active. There is precedent: `curvesHint` ("Click to add point. Drag to move. Double-click to remove.") in `packages/shared/src/i18n/en.ts:4118`, and `lassoHint` at `:1801` for the erase-object lasso. The editor's polygonal lasso has neither.

Escape to cancel an in-progress polygon is worth checking while in there. Escape is bound at `use-editor-shortcuts.ts:717` for cancelling operations; confirm it reaches this one.

## Gotchas

- Double-click must keep working. People who learned the current gesture should not lose it.
- A double-click also fires two `click` events first, so adding a click-to-close on the start vertex must not race the existing double-click path into closing twice or leaving a stray vertex.
- `finalizeLasso` (`selection-tool.tsx:489`) is the shared exit; route every new close gesture through it rather than duplicating the merge logic.
- The freehand lasso (`lasso-free`) closes on mouse-up and is unaffected. Do not change it.

## i18n

One new hint string in all 21 locale files.

DevShoaib78:
Fixes #1059.

The two gestures @CodecNomad tried in #828, Enter and clicking the first point, now close the polygon, and the options bar says so.

## What changed

**Enter closes it**, wired through `use-editor-shortcuts.ts` ahead of the crop binding.

**Clicking the first vertex closes it**, once there are at least three. The hit test is in *screen* pixels, not canvas pixels, so the target stays clickable however far the canvas is zoomed out. A small target is drawn on the first vertex as soon as the polygon can be closed, and it fills in while the pointer is over it, so the gesture is visible rather than a thing you have to know.

**The options bar carries a hint**, following the `curvesHint` / `altClickHint` precedent: "Enter, double-click or the first point closes." Added to all 21 locales so the i18n parity test stays green.

Every close gesture goes through one `closePolygon` exit into the existing `finalizeLasso`, so they all merge into the selection identically. A polygon with fewer than three vertices is discarded rather than finalised, and switching tools mid-polygon discards it, otherwise Enter could later close an invisible polygon under another tool. `lasso-free` is untouched and still closes on mouse-up, with a test pinning that.

## One thing worth your attention

I first kept `onDoubleClick` and stripped the duplicated vertex. Driving it in a real browser showed that Konva's `dblclick` fires for **any** two clicks within 400 ms, even at different positions, so placing vertices quickly would have closed the polygon under the user.

So the handler is gone and the native `MouseEvent.detail` is tested in `onMouseDown` instead: the second press of a genuine double-click closes and never adds a vertex or starts a stub, while quick single clicks at different spots stay vertices. A test covers that distinction.

## Scope

The issue also mentions checking Escape. I left it alone to keep this under your 400-line limit; it is 385 changed lines, of which 134 are the test file and 21 are the one-line hint per locale. Happy to do Escape as a follow-up.

## How to test

`vitest run tests/unit/web/polygonal-lasso-close.test.tsx` → 7 passing: closing by first vertex only once three vertices exist, the tolerance measured in screen pixels at two zoom levels, double-click closing with no duplicate vertex, quick clicks at different spots staying vertices, Enter reaching the tool through the shared holder, Enter taking priority over crop, and the freehand lasso still closing on mouse-up. I checked the two Enter cases are not vacuous by deleting the binding and confirming they fail.

`cd apps/web && pnpm typecheck` and `pnpm lint` are clean, as is `packages/shared` typecheck, and the i18n parity test passes.

Verified by hand in `apps/demo` with real mouse and key events: the hint appears with the tool active, three vertices show the target on the first point, it fills on hover, clicking it closes to a three-point lasso, double-click elsewhere closes, and Enter closes. Screenshots available if you want them here.

Not run: the root `pnpm test`, `lint` and `typecheck`, whose vitest config boots Postgres and Redis testcontainers and resolves api-only packages; I have no Docker here, so I verified per workspace for the two I touched. The full web unit suite has one unrelated failure, `tool-ui-literal-scanner`, which fails the same way on a clean checkout because it asserts on forward-slash paths and I am on Windows. Touch double-tap is unverified, though Enter and tapping the first point work there regardless.

The 20 non-English hint strings are mine rather than from `i18n:translate`, whose adapter registry does not cover the web locale files; happy to change any wording.


DevShoaib78:

DevShoaib78:
recheck

snapotter-hq:
Merged, thanks for a clean fix. One correction to the story first: on main, double-click never closed the polygon. The old `onDoubleClick` was defined but nothing wired it to the Stage, so this PR is the first working close gesture, not a preserved one. The squash message says so.

Review pass, with where each finding ended up.

**Pushed to your branch as 97507941**

- The `react-konva` Proxy mock never applied (a bare specifier from `tests/` does not resolve to the web app's `node_modules`), and if it ever had, awaiting a Proxy whose `then` never resolves would have hung collection with no error. Dropped it and the inert persist bypass; the real module loads fine under jsdom.
- "Enter beats crop" passed against an unconditional `return`. Two cases now pin that Enter still applies the crop when the holder is null or reports no polygon; both fail against that mutant.
- Tests for the tool-switch discard, the close target, and the second press of a double-click on an empty canvas. The zoom test is split so no hook stays mounted across a store reset (that was the act() warning in CI).
- The close target lights up as soon as the third vertex lands inside it, which its doc comment already claimed.
- Comments: the mouse-up branch still said only dblclick closes, the Enter hotkey never applied a transform, the double-click comment overstated the stub case, and the new banner used a double hyphen.
- es "Intro" (matches the crop string in the same namespace), ru "клик" (the file's term).

**Filed**

- #1212: Escape does not cancel an in-progress polygon. Yours if you want it.
- #1213: `setZoom` stores NaN from a zero-distance pinch. Pre-existing, found while checking the close target's zoom maths.

**Left as is, on purpose**

- A press with `detail >= 2` when nothing is in progress (a triple-click, or a quick second click at the same spot after closing) is swallowed instead of starting a polygon. Doing nothing is the safer reading of an accidental multi-click, and the next click elsewhere starts one.
- `POLY_MIN_CLOSE_POINTS` counts coordinates while "point" means a pair elsewhere in the file, and `finalizeLasso` keeps a literal 6. `PolygonCloseTarget.x/y` duplicate the first vertex. Both are naming polish for whoever touches this next.
- Unused `tolerancePx` parameter and the never-imported exports. Harmless.
- Touch: a double-tap cannot close, since a `TouchEvent` carries `detail` 0. Enter and tapping the first vertex work. The Stage has no touch-start handler at all, so that predates this PR.
- Enter from a focused element the guards do not cover (a dialog container, a `role=option`) would reach the polygon the same way it reaches crop today. Not reproduced, so not filed.

Issue #1210, opened 2026-09-20
Failed multi-file upload leaves earlier parts orphaned in object storage


snapotter-hq:
## What

The standalone upload route writes each file to object storage inside the multipart loop, before the loop finishes. If a later part is rejected and the handler returns early, the parts already written are left behind under `uploads/<jobId>/` with nothing pointing at them.

## Where it lives

`apps/api/src/routes/files.ts`, the `POST /api/v1/upload` handler: `putObject(`uploads/${jobId}/${safeName}`, safeBuffer)` runs per part inside `for await (const part of parts)`. Two early returns after a `putObject` can strand earlier objects: the SVG-sanitize rejection (added in the sanitizer-bypass PR) and, in the library route `apps/api/src/routes/user-files.ts`, the storage-quota rejection that predates it.

## Repro (expected vs actual)

1. `POST /api/v1/upload` with two parts: a valid PNG first, then an SVG over the element cap.
2. The PNG is written to `uploads/<jobId>/`, then the SVG trips the 400.
3. Expected: nothing from a failed upload persists. Actual: the PNG object remains under that `jobId` prefix, unreferenced.

Standalone uploads expire on their own schedule, so this is leaked storage, not a correctness bug. The library route additionally has a DB row per file, so a mid-loop return there can also leave a written blob with no matching row.

## Why it's out of scope

The sanitizer-bypass PR needed the rejection to fail closed (do not store the hostile file), which it does. Cleaning up siblings written earlier in the same request is a separate concern and the same gap already exists on the quota path, so fixing it in that PR would have widened it past the security change.

## Starting sketch

Buffer the parts first, or track written keys per request and delete them on any early return. A `try/finally` around the loop that removes `uploads/<jobId>/*` when the handler returns without success is the smaller change. `deleteStoredFile` / object-storage delete helpers already exist (see `apps/api/src/lib/object-storage.ts`).

## Gotchas

- The delete-on-failure path must itself not throw past the original 4xx, or a cleanup fault masks the real rejection. Mirror the fault-vs-miss handling already in the download route (`rethrowIfStorageFault`, issue #974).
- The library route spans storage plus a DB insert, so order the cleanup so a written blob without a row cannot outlive the request either.

Issue #1209, opened 2026-09-20
SVG uploads bypass sanitization via a >4KB prolog or gzip (svgz) at the storage boundary


snapotter-hq:
## What

`isSvgBuffer` decides whether an upload gets run through `sanitizeSvg`. Two inputs slip past it and get stored raw, so the sanitizer never sees them. Both predate the sanitizer-bypass fix in PR (this one) but sit in the same class it set out to close, so I'm splitting them out rather than widening that PR.

### Gap 1: the 4KB detection window

`isSvgBuffer` reads only the first 4KB (`apps/api/src/lib/svg-sanitize.ts:190`, `buffer.subarray(0, 4096)`). Pad an SVG with a leading comment or DOCTYPE longer than 4KB and the root `<svg>` falls outside the window. The prolog-strip loop then sees an unterminated comment, makes no progress, `SVG_ROOT_ELEMENT` fails, and the file is stored unsanitized. It is still a valid SVG a browser parses in full, and any downstream tool that rasterizes it (librsvg/Sharp act on content, not content-type) processes the payload.

Repro:
1. Build `<!--` + 5000 spaces + `--><svg xmlns="http://www.w3.org/2000/svg"><script>alert(1)</script></svg>`.
2. `POST /api/v1/upload` (or `/api/v1/files/upload`) with it.
3. Download it back. Expected: `<script>` stripped. Actual: stored verbatim.

### Gap 2: svgz uploads are never decompressed before storage

Both upload routes call `isSvgBuffer` on the raw buffer and never `decompressSvgz` (`apps/api/src/routes/files.ts` and `apps/api/src/routes/user-files.ts`, the SVG-sanitize block). A gzip-compressed SVG (magic `0x1f 0x8b`) is not detected as SVG and is stored compressed and unsanitized. Every other call site pairs `decompressSvgz(...)` with `sanitizeSvg(...)`, for example `apps/api/src/routes/tools/svg-to-raster.ts:233,251`. Downstream tool processing re-sanitizes, so the exposure is limited to serving the stored file directly, but the asymmetry is real.

## Why it's out of scope for the current PR

That PR fixes the `isValidImage` gate, the namespace-prefixed tag strip, and DOCTYPE/prolog handling within the existing regex sanitizer. These two are about detection completeness at the storage boundary, and the honest fix is structural rather than another regex, so they belong on their own.

## Prior art worth cribbing

[...]

PR #1211, opened 2026-09-20
fix(security): sanitize SVG uploads even when the image decoder rejects them

snapotter-hq:
Uploaded SVGs could be stored without sanitization. Both upload routes, `POST /api/v1/upload` and `POST /api/v1/files/upload`, ran the sanitizer only after a successful image validation:

```ts
const safeBuffer = isValidImage && isSvgBuffer(buffer) ? sanitizeSvg(buffer) : buffer;
```

Both routes accept any file type, so a file Sharp cannot decode is stored anyway. An SVG whose DOCTYPE declares an external entity makes librsvg fail, `isValidImage` goes false, and the raw buffer is persisted and later served as `image/svg+xml` with the payload intact. Reported by Xavier Riofrio to contact@snapotter.com.

Three defects, fixed together:

1. **The validity gate.** Both routes now sanitize on content detection alone, the way `pipeline.ts` and the tool routes already did. If `sanitizeSvg` throws (an SVG over the element cap), the route returns 400 instead of falling through to storage.
2. **Namespace-prefixed tags.** The dangerous-element strip matched only bare names, so `<x:script>` and `<svg:foreignObject>` survived on SVGs the decoder accepted. The patterns now allow an optional namespace prefix.
3. **Detection and DOCTYPE removal.** `isSvgBuffer` recognized only buffers starting with `<svg` or `<?xml`, so an SVG opening with a DOCTYPE or a comment wasn't detected at all. It now skips the XML prolog and keys off the first element, which also stops it treating an HTML document with an inline `<svg>` as SVG. The DOCTYPE regex now skips quoted strings, so a `>` or `]` inside an entity value no longer ends the match early.

Verification:

- `tests/unit/api/svg-sanitize.test.ts`: prefixed tags, bracket-in-quote DOCTYPE, prolog-aware detection, and the HTML-with-inline-svg negative. All three sanitizer unit suites pass (149 tests).
- `tests/integration/security/svg-upload-sanitization.test.ts`: uploads each payload class to both routes, downloads it back, asserts the stored bytes are sanitized. The adversarial, library, and upload-auth integration suites pass alongside it.
- `pnpm typecheck`, Biome, and `pnpm check:license-boundary` are clean.

Downloads already carry `Content-Disposition: attachment` and `X-Content-Type-Options: nosniff`, and the CSP has no `unsafe-inline` for scripts, so a stored SVG doesn't execute as a document in the app origin. This PR closes the reported gate bypass and the two sibling sanitizer gaps.

Two residual gaps stay open, both pre-existing and out of scope here, filed as follow-ups:

- #1209: `isSvgBuffer`'s 4KB detection window and unhandled svgz let a crafted SVG reach storage undetected. The durable fix the sanitizer comment already flags (XML parse plus an element allowlist) subsumes both.
- #1210: a multi-file upload that rejects mid-loop leaves earlier parts orphaned in object storage.

Reviewed before opening with the silent-failure-hunter and pr-test-analyzer agents; their findings and how each was handled are in a PR comment.


snapotter-hq:
Pre-merge review ran the silent-failure-hunter and pr-test-analyzer agents on the diff. What they found and what changed:

**Swallowed sanitizer errors (silent-failure, the important one).** My first cut wrapped `sanitizeSvg` in a bare `catch {}` that returned a generic 400 and discarded the error. That masks the sanitizer's own actionable messages (size cap, element cap) and, worse, would bury a genuine server fault (a `RangeError` from `toString` on a huge buffer under an unlimited size cap) as a bad-upload 400 with nothing in the logs or Sentry. Fixed: the sanitizer's limit guards now throw with `statusCode: 400`, and both routes return that reason for a known rejection but log and `reportError` + 500 for anything unexpected. Throwing straight out of the multipart loop gets rewritten to a generic 400 by the stream teardown, which is why this is handled in the loop the way the adjacent quota check already is, rather than left to propagate.

**The 400 branch was untested.** Added an integration case that uploads an SVG over the 5000-element cap to both routes and asserts a 400 naming the reason, with nothing persisted.

**The XXE regression test leaned on librsvg behavior it never asserted.** The gate-removal fix only has teeth because librsvg rejects the undefined entity. If a future libvips tolerated it, the test would silently degrade to "sanitizer ran anyway" and still pass. Added an assertion that `validateImageBuffer` rejects the XXE fixture, pinning the premise.

**Sanitizer edge cases the new tests didn't cover.** Added unit cases for a `>` inside a quoted DOCTYPE external id, a single-quoted DOCTYPE value, a mismatched-prefix tag pair (`<x:script>...</script>`), a prefixed `<x:iframe>`, and the `<svgother>` negative that guards `isSvgBuffer` against going over-broad.

**Considered and deferred, not fixed here:**

- `isSvgBuffer`'s 4KB detection window and unhandled svgz let a crafted SVG reach storage undetected. Pre-existing, and the honest fix is the XML-parse allowlist the sanitizer comment already flags, so it goes to #1209 rather than another regex here.
- A multi-file upload that rejects mid-loop orphans earlier parts in object storage. Pre-existing (the quota path has the same shape) and a separate concern, filed as #1210.

Full suite green: the three sanitizer unit suites, the new upload-sanitization integration spec, and the adversarial, library, and upload-auth suites (234 tests). Typecheck, Biome, and the license-boundary check pass.

Issue #1217, opened 2026-09-20
fix(import): the row-count guard cannot detect a dropped row under --force


snapotter-hq:
The 1.x importer checks it copied everything by comparing the target table's total row count against the number of source rows. On an empty target that works. Under `--force` the target already has rows, so the total is inflated past `rows.length` and the check passes no matter how many rows went missing.

## Repro

1. Import a 1.x file into a fresh instance so the target has rows.
2. Run `migrate:sqlite -- <another 1.x file> --force` where that second file's rows would not all land.

Expected: a row-count mismatch error.

Actual: `count` is the whole table, so `count < rows.length` is false and the check passes. `result.tables[table]` is then set to `rows.length`, the source count, and reported as though it were the imported count.

## Where it lives

`apps/api/src/db/migrate-from-sqlite.ts:323-329`:

```ts
const count = (
  await tx.execute(sql`SELECT count(*)::int AS n FROM ${sql.raw(`"${table}"`)}`)
).rows[0].n as number;
if (count < rows.length) {
  throw new Error(`Row count mismatch for ${table}: sqlite=${rows.length} pg=${count}`);
}
result.tables[table] = rows.length;
```

## Why it was out of scope for #1005

Pre-existing, and #1005 did not make it worse: that fix inserts every source row and only nulls a column. It is worth writing down now because #1005 makes `--force` the sanctioned route for absorbing identity collisions, so a guard that is inert on that path now covers ground it did not before.

## Gotchas

Every INSERT in the loop is unconditional, so today there is no path that drops a row silently. This is a guard that would not catch one if a future change introduced it, not a live bug. Worth keeping the fix proportionate.

## Starting sketch

Count rows before the table's insert loop and compare the delta, or keep an explicit counter incremented per successful INSERT, and compare that against `rows.length`. Set `result.tables[table]` from the same number so the reported count means "rows this import inserted" on both paths. `tests/integration/platform/migrate-from-sqlite.test.ts` has `--force` fixtures in the "an identity the target already holds" suite to build on.

Issue #1216, opened 2026-09-20
fix(import): a login racing a --force import still surfaces a bare 23505


snapotter-hq:
#1005 settles duplicated SSO identities by pre-scanning the source and comparing against the identities the target already holds. That read happens once, before any INSERT. An OIDC login that commits a new identity between the read and the INSERT is invisible to it, so the INSERT still raises 23505 and the operator still gets the bare driver error #1005 set out to remove.

## Repro

1. Run an instance with OIDC configured and at least one user.
2. Start `pnpm --filter @snapotter/api migrate:sqlite -- <1.x file> --force` where the 1.x file carries a user with `(auth_provider, external_id)` that nobody holds yet.
3. Between the `heldIdentities` read and the users INSERT, have someone complete a first OIDC login for that same external identity.

Expected: an error naming the identity that lost, and saying a concurrent login took it.

Actual: `duplicate key value violates unique constraint "users_auth_provider_external_id_unique"` surfaced raw, the whole import rolled back. On the boot path that becomes `FATAL: 1.x SQLite import failed ...` and `process.exit(1)` at `apps/api/src/db/sqlite-import.ts:175-178`.

## Where it lives

`apps/api/src/db/migrate-from-sqlite.ts`: `heldIdentities(tx)` is awaited once at `:298`; the INSERT that can collide runs at `:319-321`. The advisory lock at `:268` serializes concurrent *imports*, not concurrent logins, and the transaction is READ COMMITTED, so a row committed by another session after the snapshot read is invisible to `taken` but visible to the unique index.

## Why it was out of scope for #1005

That issue's stated fix was the pre-scan, and its Gotchas explicitly ruled out per-row 23505 catching as the dedupe mechanism. This is the residual race the pre-scan cannot close, and how to word and handle it is a separate decision.

## Gotchas

- Do not swallow 23505 around the users INSERT. It also covers `users_username_unique` and the primary key, and hiding those is exactly the silent failure this codebase cares about. Translate, then rethrow.
- Narrow window in practice: the boot path only imports into a target with zero users, so it needs a sibling replica serving logins into an empty instance. The realistic case is a `--force` CLI run against a live instance.

## Starting sketch

[...]

Issue #1215, opened 2026-09-20
feat(import): let migrate:sqlite --dry-run preview duplicated SSO identities


snapotter-hq:
`migrate:sqlite --dry-run` exists to tell an operator what the import will change before it writes anything, and the upgrade guide sells it that way. Since #1005 the import has a second lossy conversion, detaching a duplicated SSO identity, and the dry run says nothing about it. The operator reads `would import 4 rows total`, reasonably concludes nothing surprising is coming, and finds out from stderr after the write has committed.

## Repro

1. Build a 1.x SQLite file whose `users` table holds two rows for one `(auth_provider, external_id)`.
2. `pnpm --filter @snapotter/api migrate:sqlite -- <path> --dry-run`

Expected: a line in the same spirit as the existing job-status line, e.g. `1 external identity is shared by 2 users; only the oldest row keeps it`.

Actual: row counts, blob counts, job statuses to be mapped, and nothing about identities.

## Where it lives

- `apps/api/src/db/sqlite-import.ts:92-123`: `analyzeSqlite` builds `AnalyzeReport`. It already reports the *other* lossy conversion, out-of-enum job statuses, at `:108-117`.
- `apps/api/src/db/sqlite-import.ts:204-225`: the dry-run print block, including `job statuses to be mapped: ... -> failed/processing` at `:214-217`.
- `apps/docs/guide/upgrading.md:82` describes what the dry run prints: "row counts per table, how many saved-library files it found on disk, and any job statuses it will normalize."

## Why it was out of scope for #1005

That issue scoped the fix to the copy itself and its per-row log line. The dry run is a different function in a different file with a hard constraint of its own (see below), and the docs change reaches 21 locales, so bundling it would have made the fix hard to review.

## Gotchas

- `analyzeSqlite` must work with no live Postgres, which is the whole point of it: operators run it before starting the stack. Intra-source twins are detectable from the SQLite file alone with `SELECT auth_provider, external_id, count(*) FROM users WHERE external_id IS NOT NULL GROUP BY 1, 2 HAVING count(*) > 1`, so that half is fine.
- The other half, an identity the *target* already holds, genuinely needs a live target and cannot be previewed. A partial preview still beats none; just do not word it as though it were complete.
- `apps/docs/guide/upgrading.md` is translated into 21 locales. Per CLAUDE.md the web i18n pipeline is a three-step handoff, not a one-shot run, so budget for that rather than editing the English file alone.

## Starting sketch

[...]

Issue #1214, opened 2026-09-20
fix(import): a detached SSO identity leaves no durable record of what the import changed


snapotter-hq:
The 1.x import can now unlink a user's SSO identity (#1005), and the only trace is a `console.warn` line. On the boot path, which is how most self-hosted upgrades actually run, that line is one entry in container stdout at first boot and it is gone after the first log rotation. Three weeks later the operator gets "Bob can't sign in with SSO any more" and there is nothing in the database to correlate against.

## Repro

1. Build a 1.x SQLite file whose `users` table holds two rows for one `(auth_provider, external_id)` (the `buildTwinSource` helper in `tests/integration/platform/migrate-from-sqlite.test.ts` builds exactly this).
2. Boot a fresh instance with `SQLITE_MIGRATE_PATH` pointing at it.
3. Read back the import marker: `SELECT value FROM settings WHERE key = 'sqlite_import'`.

Expected: the marker records that an identity was detached, and which row.

Actual: the marker holds `status`, `tables`, `blobs`, `sourcePath`, `at` and nothing else. `Imported 1.x SQLite database: {...}` on stdout reads as a complete summary of what the import did, and it omits the one lossy thing it did.

## Where it lives

- `apps/api/src/db/migrate-from-sqlite.ts:335-337`: the warn loop. The `detached` array is local and never reaches `MigrationResult` (`:9-11`), so no caller can see it.
- `apps/api/src/db/sqlite-import.ts:158-169`: `runBootImport` writes the marker and logs the success line, neither of which mentions detachments.
- `apps/api/src/db/sqlite-import.ts:228-231`: the CLI prints `Migration complete:` on stdout *after* the warnings go to stderr, so `migrate:sqlite -- db.sqlite > import.log` keeps the all-clear and throws the warnings away. Exit code is 0 either way.

## Why it was out of scope for #1005

That issue's own sketch scoped the reporting to "log one line per detached row", which is what shipped. Putting the list in `MigrationResult` changes the marker's stored shape, so it wants its own change and its own thought about what a reader three weeks later needs.

## Prior art

`apps/api/drizzle/0008_users_identity_unique.sql:7-9` has the same problem and documents a recovery query in a comment, because a migration cannot write the audit log (rows there carry an HMAC). The importer is application code with the HMAC key available, so `auditLog` in `apps/api/src/lib/audit.ts:47` is reachable after the transaction commits. Worth considering alongside the marker.

## Gotchas

[...]

PR #1218, opened 2026-09-20
fix(import): settle twin SSO identities during the 1.x SQLite import

### LINKED ISSUE #1005: fix(import): 1.x SQLite import fails outright when the source holds twin SSO identities
Migration 0008 (the #969 fix) adds `users_auth_provider_external_id_unique`, a partial unique index on `(auth_provider, external_id)`. The 1.x SQLite import copies `users` rows verbatim, and 1.x had the same auto-create race, so a source database can hold two rows for one identity. Those rows now trip the index and the whole import rolls back.

## Repro

1. Build a 1.x SQLite file whose `users` table has two rows with `auth_provider = 'oidc'` and the same `external_id` (the `buildFixtureSqlite` helper in `tests/integration/platform/migrate-from-sqlite.test.ts` is the easiest starting point).
2. `pnpm --filter @snapotter/api migrate:sqlite -- <path>`

Expected: the import completes, handling the twin the way migration 0008 handles it on an upgraded Postgres install (oldest row keeps the identity, the rest get `external_id = NULL`).

Actual: the second row's INSERT raises 23505 on `users_auth_provider_external_id_unique`, the transaction rolls back, and the operator sees a raw driver error that doesn't name the duplicate.

## Where it lives

`apps/api/src/db/migrate-from-sqlite.ts`, `migrateFromSqlite()`: `runMigrations()` runs first, so the index exists, then the per-table loop inserts each converted row with `tx.execute(sql\`INSERT INTO ...\`)`. There is no per-row conflict handling; any failure aborts the transaction.

## Why it was out of scope for #969

That fix covers the live login race and the upgrade path for Postgres installs. The import is a one-shot 1.x to 2.x path with its own error contract (`TargetNonEmptyError`, the row-count mismatch check), and how to report a detached twin to the operator deserves its own change.

## Prior art

- `apps/api/drizzle/0008_users_identity_unique.sql`: the keep-rule (oldest `created_at`, then smallest `id`) and the detach (`external_id = NULL`). The import should apply the same rule so both paths agree.
- `apps/api/drizzle/0007_needy_fenris.sql`: the same shape for case-twin team names.

## Gotchas

- The import inserts inside one transaction, so the dedupe has to happen before the INSERTs (pre-scan the SQLite rows), not by catching 23505 per row.
- 1.x stored `created_at` as integer seconds; `convertRow` converts it, so compare converted values.
- `--force` inserts into a non-empty target, where the older row may already live in Postgres. The keep-rule then has to look at both sides.

## Starting sketch

In `migrateFromSqlite`, after reading the `users` rows and before inserting: group by `(auth_provider, external_id)` where `external_id` is not null, sort each group by converted `created_at` then `id`, and set `external_id = null` on every row but the first. Log one line per detached row (username, provider, external id) so the operator can find them afterwards. Add a case to `tests/integration/platform/migrate-from-sqlite.test.ts` with two twin rows and assert that both import, one keeps the identity, and the log names the other.

snapotter-hq:
A 1.x source holding two `users` rows for one external identity took the whole import down.

`runMigrations()` runs before the copy, so migration 0008's partial unique index on `(auth_provider, external_id)` already exists when the rows land, and 1.x carried the same auto-create race 0008 was written to clean up. The second INSERT raised 23505, the transaction rolled back, and the operator got a driver error that never named the duplicate:

```
duplicate key value violates unique constraint "users_auth_provider_external_id_unique"
detail: Key (auth_provider, external_id)=(oidc, ext-twin) already exists.
```

Reproduced against unmodified `main` before any code changed, at `apps/api/src/db/migrate-from-sqlite.ts:212`.

## What changed

`users` now converts up front and settles the twins before any INSERT, so the copy never offers the index a duplicate. Every other table stays row-at-a-time; converting all of them would hold a large 1.x `jobs` table twice over for no benefit, since `users` is the only table here with a cross-row constraint to settle.

Inside the source the keep-rule is 0008's, so an imported install lands in the shape an upgraded one does: oldest `created_at` keeps the identity, ties break on the smallest id, the rest get `external_id = NULL`. Ties are the normal case rather than the exotic one, because 1.x stores epoch seconds and the race fires well inside a second.

Each detached row is reported once the transaction commits, naming the row that kept the link. Unpicking this afterwards means hand-written SQL against `external_id` (there is no UI for it), so the winner is the one fact an operator actually needs.

## The deliberate deviation from 0008

For an identity the **target** already holds, only reachable with `--force`, the incoming row always loses, even when it is older. That is not 0008's oldest-wins, and it is on purpose.

Across two independent installs `created_at` is not a comparison of anything: the 1.x row's timestamp is a fact about a system that no longer runs. The harm is asymmetric too. Detaching the target row locks out someone who signs in today, with no password to fall back on, since an SSO row carries `password_hash` NULL, and admins are the people most likely to have been SSO-provisioned. Detaching the incoming row inconveniences an account that has never logged into this install. Both need hand repair; only one is an incident.

`TargetNonEmptyError` now says so, which is the message an operator reads before deciding to pass `--force`.

## Verified

Ran locally, all in this session:

- `pnpm vitest run` on the migrator, its caller (`sqlite-import-orchestrator`), the schema-drift guard and `migrate-sqlite-permissions`: 54 passed.
- Teeth: reverted the production file with the tests in place, 19 failed; restored, 36 passed. A targeted mutant moving the warn loop inside the transaction is killed by "reports nothing when the import rolls back".
- `pnpm lint` and `pnpm typecheck` green, at the same 573 warnings / 3 infos as the pre-change baseline.
- Blast radius: `migrateFromSqlite` has two callers, `runBootImport` and the CLI, both in `sqlite-import.ts`. `TargetNonEmptyError` is matched by class, not message, so the longer text is safe. Ran the suites covering both.
- Full-suite baseline diff: recorded in a comment below.

## Review findings

Three pr-review-toolkit agents ran against `28f16a8e`. code-reviewer found no critical or important defects.

Fixed here, in `8af7a1c6`:

- The warning said "another account already answers to it", false for the headline case where both twins arrive in the same import. The two cases now get different text, and the winner is named.
- `identityKey` treated a missing `auth_provider` column as `""` while Postgres fills the column default `'local'`, so a pre-1.17.2 source under `--force` keyed its rows differently from the target's, missed the collision and hit the same 23505 this fixes.
- The `--force` case moved into its own suite with its own fixtures; it had depended on declaration order and would have failed under a vitest retry.
- Twin fixtures are built on the real 1.17.2 schema replayed from the archived legacy migrations, instead of a fourth hand-written copy of the `users` DDL.
- Added coverage for a rolled-back import reporting nothing, a three-row group, a held identity under a different provider, and a source predating `auth_provider`.

Considered and dismissed:

- Guarding `oldestFirst` against a null or non-numeric `created_at`. Both reviewers traced it independently: the column is NOT NULL in the target with no DB default, so either case fails the INSERT and rolls back regardless. Unreachable in an import that survives, so it stays uncoded.
- Removing the null check in `heldIdentities`, unreachable because the query filters `external_id IS NOT NULL`. Keeping it lets `identityKey` stay the single definition of the key, and the SQL filter is what lets the partial index serve the query.

Filed rather than fixed, all out of scope for this issue: #1214 (no durable record of a detachment), #1215 (dry run cannot preview twins), #1216 (a login racing a `--force` import still surfaces a bare 23505), #1217 (the row-count guard is inert under `--force`, pre-existing).

Closes #1005


deepsafehq:
## Baseline diff

Two full `pnpm test` runs at `VITEST_MAX_FORKS=2`, sequential, nothing else on the box: the unfixed tree in a detached worktree at the base commit, then this branch.

| | files | tests | failures |
|---|---|---|---|
| baseline (unfixed, f26db6d6) | 897 | 21187 | 1 |
| this branch | 898 | 21207 | 1 |

`comm` on the two failure lists:

```
NEW (post-change only):
  × SCIM ... Groups CRUD > concurrent PUT renames onto one displayName return one 200 and one SCIM 409, never 500
BASELINE ONLY:
  × SCIM ... Groups CRUD > concurrent duplicate group creates return one 201 and one SCIM 409, never 500
IN BOTH: (none)
```

One in, one out, both in `scim.test.ts`, neither anywhere near this change. That is the tracked licence-gate race (#1156, #980, #990, #1000), so I checked the signature rather than assuming:

- baseline: `only 1/2 contenders blocked on insert into "teams"; contenders finished without blocking (outcome: 201,403)`
- this branch: `only 1/2 contenders blocked on update "teams"; contenders finished without blocking (outcome: 200,403)`

Both are the 403 contender from the licence gate, which is the documented shape, and it lands on a different test each run. Then, one run each, isolated:

- `scim.test.ts` alone on this branch: 88 passed.
- `scim.test.ts` alone on the unfixed baseline worktree: 88 passed.

So it reproduces only under full-suite load, on both trees. Pre-existing, not refiled.

The +1 file and +20 tests are all from this PR: 13 in the new unit file, 7 added to the migrator integration suite.

## Review findings

Three pr-review-toolkit agents ran against 28f16a8e. What each finding did, since none of it is visible outside the session:

Fixed in 8af7a1c6:

1. The warning claimed "another account already answers to it", which is false for the headline case where both twins arrive in the same import, and it named only the row that lost. `DetachedIdentity` now carries the winner and the two cases get different text.
2. `identityKey` keyed a missing `auth_provider` column as `""` while Postgres fills the column default `'local'`. A pre-1.17.2 source under `--force` would have missed the collision and hit the same 23505 this PR exists to fix.
3. The `--force` test depended on declaration order and would have failed under a vitest retry by re-importing rows it had already inserted. It moved to its own suite with its own fixtures.
4. The twin fixture was a fourth hand-written copy of the 1.x `users` DDL, so it tested the fixture's idea of 1.x rather than the real thing. It now replays the archived legacy migrations via `buildLegacySqlite`.
5. Missing coverage added: a rolled-back import reporting nothing, a three-row group, a held identity under a different provider, and a source predating `auth_provider`.

Considered and dismissed:

- Guarding `oldestFirst` against a null or non-numeric `created_at`. Two reviewers traced it independently and agreed: the target column is NOT NULL with no DB default, so both cases fail the INSERT and roll back regardless. Unreachable in an import that survives.
- Dropping the null check in `heldIdentities`, unreachable because the query filters `external_id IS NOT NULL`. Keeping it lets `identityKey` stay the single definition of the key format, and the SQL filter is what lets the partial index serve the query.
- Rewriting the insert loop's index-plus-ternary into `rows.entries()`. Cosmetic.

Filed instead of fixed, all outside what #1005 asked for: #1214, #1215, #1216, #1217.

## Teeth

- Reverted the production file with the tests in place: 19 failed. Restored: 36 passed.
- Targeted mutant moving the warn loop inside the transaction: killed by "reports nothing when the import rolls back".

Issue #1220, opened 2026-09-20
fix(upload): auto-orient re-encodes rotated photos at Sharp's default quality


snapotter-hq:
Every photo with an EXIF orientation tag gets re-encoded before any tool touches it, at Sharp's default quality rather than the 95 the rest of the pipeline uses. Phone photos carry that tag constantly, so this is most uploads from a camera roll.

## Where

`apps/api/src/lib/auto-orient.ts:19`

```ts
export async function autoOrient(buffer: Buffer): Promise<Buffer> {
  try {
    const meta = await sharp(buffer).metadata();
    if (meta.orientation && meta.orientation > 1) {
      return await sharp(buffer).rotate().toBuffer();
    }
```

The `toBuffer()` names no format, so Sharp writes it back in whatever container came in. For a JPEG that is a second lossy generation at the encoder's own default, inserted before the tool the user actually asked for has run.

## Repro

Upload any JPEG whose EXIF orientation is 2 to 8 (anything shot in portrait on a phone) and compare the buffer `autoOrient` returns against a rotate that names its quality. The gate is `meta.orientation > 1`, so an already-upright photo is untouched and only rotated ones pay it.

## Why it was out of scope

Found while fixing #1190, which is the same class: an unnamed intermediate written back in the input's container. Every other instance took `.png()`, and that is the wrong answer here. `autoOrient` runs on most upload paths and hands its result to the tool, so turning every rotated photo into a PNG would inflate what the rest of the pipeline carries, on a stack that documents a 2GB floor.

The GIF half of #1190 does not apply, because GIFs carry no EXIF orientation and never reach the branch.

## Fix sketch

Keep the container and name the quality, rather than letting Sharp pick:

```ts
const detected = await resolveOutputFormat(buffer, "");
return await sharp(buffer).rotate().toFormat(detected.format, detected.encoderOptions).toBuffer();
```

`resolveOutputFormat` already maps a detected format to an encoder plus a quality of 95, and since #1190 its `encoderOptions` carries the GIF palette rule too, so going through it avoids re-deriving either.

Watch the formats that have no Sharp encoder. `resolveOutputFormat` falls back to PNG for SVG, RAW, PSD and friends, which is right for a tool's output and wrong here: a RAW file that somehow reached this branch would come out PNG. None of those carry EXIF orientation today, so the branch is unreachable for them, but a fix should not depend on that staying true.

## Guard

[...]

Issue #1224, opened 2026-09-20
fix(image-enhancement): a failed deep-enhance pass returns 200 with no trace


snapotter-hq:
A user ticks Deep Enhance, the SCUNet pass fails, and they get a 200 with the shallow result. Nothing in the response, the logs or Sentry says the feature they asked for did not run.

## Where

`apps/api/src/routes/tools/image-enhancement.ts`, the deep-enhance block:

```ts
if (settings.deepEnhance && isToolInstalled("noise-removal")) {
  ...
  try {
    const result = await noiseRemoval(buffer, scratchDir, { ... });
    buffer = result.buffer;
  } catch {
    // SCUNet unavailable, fall back to the Sharp-only result
  } finally {
```

The catch is empty and unconditional. A sidecar crash, an OOM, a bad scratch dir and a genuinely missing model all land in the same place and all look like success from outside.

`isToolInstalled` gates the branch, so the comment's "SCUNet unavailable" case is already handled before the try. What actually reaches this catch is a pass that was supposed to run and broke.

## Expected vs actual

Expected: either the deep result, or a signal that it was asked for and did not happen.
Actual: the shallow result, a 200, and silence.

## Fix sketch

The fallback is right; the silence is not. At minimum log it with the error attached, so it is greppable and reaches Sentry:

```ts
} catch (err) {
  request.log.warn({ err }, "deep enhance failed, returning the Sharp-only result");
}
```

Whether the response should say so is a separate call. The tool has no field for partial success today, and adding one touches the frontend and the i18n catalogue, so the log is the part that can land on its own.

Worth checking the same pattern in the other AI routes while in here; this is the one that turned up under #1190.

## Out of scope because

Pre-existing, and #1190 was about container handling. It sits about thirty lines below a hunk that PR edited, which is how it was spotted. CLAUDE.md treats no-silent-failures as a house rule, so it gets its own issue rather than a drive-by fix in an unrelated PR.

Issue #1223, opened 2026-09-20
fix(gif-tools): the optimize colour count has four effective settings, not 255


snapotter-hq:
The optimize mode's colour slider takes anything from 2 to 256 and has four settings that do anything. Ask for 32 on a 256-colour GIF and you get 254 colours back and a byte-for-byte identical file.

## Repro

`tests/fixtures/image/formats/sample.gif`, 256 colours, 194757 bytes, straight through Sharp's encoder as the route calls it:

```
requested   4 ->   4 distinct,  27268 bytes
requested   8 ->  15 distinct,  58020 bytes
requested  16 ->  15 distinct,  58020 bytes
requested  32 -> 254 distinct, 194027 bytes
requested  64 -> 254 distinct, 194027 bytes
requested 128 -> 254 distinct, 194027 bytes
requested 256 -> 254 distinct, 194027 bytes
```

Same with `reuse: false`, so this is not #1190.

## Cause

libvips writes GIF at a power-of-two bit depth and rounds `colours` up to reach one. 4 gives 2 bits, 8 and 16 both give 4, and anything from 17 up gives 8 bits, which is 256 and therefore no reduction on a source that already fits. The slider in `apps/api/src/routes/tools/gif-tools.ts` (`colors`, min 2, max 256, default 256) offers 255 values on top of that.

A user dragging it to 64 to shrink a file gets nothing and no explanation.

## Fix sketch

Two ways, and the choice is a product one.

Snap the control to what the encoder can do: 2, 4, 16, 256, presented as a small set rather than a slider. Honest and cheap, and it makes the file-size effect predictable.

Or quantise before encoding, so the requested count is real. That means doing the palette reduction ourselves rather than asking libvips for a bit depth, and it is the only way an intermediate value like 64 can mean anything.

Whichever way it goes, the frontend control in `apps/web/src/components/tools/gif-tools-settings.tsx` and the i18n strings for it move with the schema, and new strings go in all 21 locale files.

## Gotchas

- Measured per distinct colour in the decoded output, not from the palette table, because the encoder writes a table sized to the bit depth whether or not the entries are used.
- The 8 and 16 rows landing on 15 rather than 16 is dithering spending one entry, not an off-by-one.

## Out of scope because

Found while checking whether optimize mode needed the palette change in #1190. It does not: the mode behaves the same with reuse on or off. This is a separate question about what the setting promises.

Issue #1222, opened 2026-09-20
fix(image-engine): GIF encodes never ask for a fresh palette


snapotter-hq:
#1190 fixed GIF palette reuse across `apps/api`, where every route now resolves its encoder options through `resolveOutputFormat`. `packages/image-engine` encodes GIFs on three of its own paths and none of them went through that, so the bug is still live wherever a pipeline ends in a GIF.

## Where

- `packages/image-engine/src/engine.ts:113`: `image.toFormat(sharpFormat as SharpFormat)` with no options at all. `processImage` runs arbitrary named operations first, so anything that resizes, blurs or adjusts before a GIF output reintroduces #1190 exactly.
- `packages/image-engine/src/operations/convert.ts:26-34`: `FORMAT_MAP` includes gif, and the pipeline handed in may already be resized.
- `packages/image-engine/src/operations/compress.ts:17-21` (`formatOpts`) and its callers at `:59`, `:87`, `:110`, `:140`. `compressToTargetSize` resizes and re-encodes at `:140`, which is the resample-then-reuse case.

## Repro

Through `compress` in target-size mode, on a GIF small enough that the quality sweep has to fall back to resizing:

```
compress with resize: gif(default) vs the same pixels as png -> worst channel 9
                      gif(reuse: false)                      -> 0
```

Quality mode never resizes, which is why the sweep in `tests/unit/api/image-container-independence.test.ts` runs `compress` with `{}` and sees nothing.

## Second thing in the same file

Sharp's GIF encoder has no `quality` parameter. `toFormat("gif", { quality: 95 })` is byte-identical to `toFormat("gif")`. `findBestQuality` (`compress.ts:70-98`) binary-searches quality anyway, so compressing a GIF runs roughly seven encodes that all produce the same bytes before falling through to the resize path. Worth an early branch for GIF that goes straight to resizing.

## Fix sketch

`resolveOutputFormat` lives in `apps/api/src/lib/output-format.ts` and `packages/image-engine` must not import from `apps/*` (`pnpm check:license-boundary` enforces it), so the rule needs its own home here rather than an import. `formatOpts` in `compress.ts` is the closest thing to a shared option builder already; the smallest version is to add `reuse: false` for gif there and give `engine.ts` and `convert.ts` the same treatment.

If it grows past that, the honest version is one exported helper in `packages/image-engine` that both sides use, with `apps/api`'s `encoderOptions` built on top, so the rule is written once.

## Gotchas

[...]

PR #1226, opened 2026-09-20
fix(image): answer by pixels, not by the container they arrived in

### LINKED ISSUE #1190: fix(image): unnamed intermediate buffers re-encode in the input's container, corrupting still GIFs

### LINKED ISSUE #1219: fix(image-enhancement): corrections have no effect on an opaque still GIF because the encoder reuses the input's palette
Corrections applied by image-enhancement have no visible effect on an opaque still GIF. The output decodes fine and returns 200, but the pixels come back close to the uncorrected input.

## Repro

A two-tone opaque GIF (left half `120,20,20`, right half `20,20,120`, no alpha at all) through a brightness boost (`linear(1.8, 0)`, standing in for a real correction):

```
PNG in -> corrected -> left pixel  216,36,36   (brightened, correct)
GIF in -> corrected -> left pixel  120,20,20   (unchanged from input)
```

The GIF output is pixel-identical to what went in. The correction ran, and the result was thrown away on the way out.

## Cause

`apps/api/src/routes/tools/image-enhancement.ts`, the no-alpha branch of `core()` (currently the `else` in the `if (alphaBuffer)` split added by #1187):

```ts
buffer = await image.toFormat(outputFormat.format, { quality: outputFormat.quality }).toBuffer();
```

When the input is a GIF, `image` is a Sharp pipeline that originally read a GIF, and Sharp's GIF writer defaults to `reuse: true`: it reuses the palette attached at read time, not one built from the corrected pixel data. `applyCorrections()` changes the actual color values, but the encoder quantizes those new values against the stale, pre-correction palette, so any color that isn't already in that palette snaps back to the nearest match, usually the original color itself for a low-color-count image.

This is the same `reuse: true` mechanism behind #1180 and #1190, but a different shape: those are about a bare `.toBuffer()` silently defaulting to the input's container. Here the format is explicitly named (`outputFormat.format`, correctly "gif"), and it still corrupts, because naming the target format doesn't opt out of palette reuse.

## Where else

The has-alpha branch of `core()` accidentally dodges this as a side effect of #1187's fix: it now routes the corrected image through a named PNG intermediate before the final GIF encode, which breaks the "read from GIF" lineage the reuse defaults key off. The no-alpha branch was left as-is since #1187 was scoped to the alpha symptom, so it's still exposed. Worth checking whether any of the sixteen call sites #1190 lists have the same "explicitly named but still GIF-sourced" shape rather than the bare-`toBuffer()` shape it scanned for.

## Fix sketch

Mirror #1187's fix in the no-alpha branch too: materialize the corrected image as a named PNG buffer, then encode that (not the live pipeline) to `outputFormat.format`:

```ts
buffer = await sharp(await image.png().toBuffer())
  .toFormat(outputFormat.format, { quality: outputFormat.quality })
  .toBuffer();
```

`.gif({ reuse: false })` on the direct encode is a narrower alternative, but `toFormat()` doesn't take a `reuse` option (it's GIF-specific), so it would need a conditional keyed on `outputFormat.format === "gif"`. The PNG-intermediate approach handles every format uniformly.

## Gotchas

- Only still GIFs are exposed. An animated GIF is split into per-frame PNGs before reaching `core()` (#1083), so it never carries the read-from-GIF lineage into a correction step.
- A regression test here can't just check status or dimensions; it needs to compare corrected pixel values against a PNG-input reference, the way `tests/unit/api/duotone-container-independence.test.ts` does for #1180.
- Any of `packages/image-engine`'s other GIF-touching call sites that name their target format explicitly are worth the same check; #1190's grep only catches the bare-`toBuffer()` shape.

## Out of scope because

Found while implementing #1187 (transparent GIF loses alpha). That issue was scoped to the alpha/transparency symptom; this is a broader, alpha-independent bug in the same function that needs its own test and its own look at whether the fix should be shared between the two branches.

snapotter-hq:
Closes #1190
Closes #1219

Put a green border on a GIF and it came back `rgb(76,105,113)`, a muddy grey-blue, with a 200 and no warning. Sharpen one and the halos came back in the source's own colours. Enhance one and the corrections had no visible effect at all. Convert a transparent one to PDF and the page behind it was that same grey-blue instead of white.

## Why

Sharp's GIF writer defaults to `reuse: true`. It re-expresses the image in the palette libvips read the input with, so a colour the tool just introduced has nowhere to go. At every palette size from 2 entries to 200:

```
extend with #008000 on a GIF:  reuse -> 0,0,0     reuse:false -> 0,128,0
sharpen a 4-colour GIF:        reuse -> 61 off    reuse:false -> 0 off
```

Two code shapes walked into it, and only the first is the one #1190 describes.

**An unnamed intermediate.** `await sharp(buf).extend(...).toBuffer()` names no format, so Sharp picks the input's container and the damage lands mid-pipeline, before the next step sees the pixels. On a JPEG the same line inserts a lossy generation at the encoder's own default, in a pipeline the caller asked to run at 95.

**The final encode.** Tools with no intermediate at all were corrupted just the same. `rotate` at 33 degrees was 111 off, `color-blindness` 127, `adjust-colors` with sharpness 32, `sharpening` and `resize` likewise. Naming intermediates cannot help these; there is nothing to name.

#1219 is the second shape, filed separately this morning. Its fix sketch assumes `toFormat()` cannot carry `reuse` because the option is GIF-specific. It can: every other encoder ignores keys it does not know, and I checked that on png, jpeg, webp, tiff and avif before building on it. That is what makes one central fix possible instead of a conditional per call site.

## What changed

`resolveOutputFormat()` now returns the whole `toFormat()` option bag as `encoderOptions`, carrying `reuse: false` for GIF. Routes pass it through instead of rebuilding `{ quality }` by hand. For every non-GIF format the bag is exactly `{ quality }`, so nothing outside GIF output changes behaviour. `outputFormatFor()` covers the two routes that pick their own format, which previously wrote the quality out twice and could drift.

Every intermediate that named no format now names PNG: border, image-pad, vectorize, vignette, circle-crop, rounded-crop, image-to-pdf, image-enhancement, and the multi-file and model-backed ones the issue called a floor rather than an inventory (stitch x9, collage, sprite-sheet, compose x2, smart-crop x7, passport-photo x3, transparency-fixer).

Five more live in `lib/bg-effects.ts`, reached by blur-background, remove-background and background-replace. They are not in #1190's list and were not in my first scan either, because I pointed it at `routes/tools` and they sit in `lib`. Same signature, same 76,105,113.

#1187 landed separately in #1225 while this was in review, and reached the same two-part fix I had: name the alpha lift, and name the colour pass before `joinChannel`, because a GIF decodes to four channels and the mask was being appended as a fifth band nothing reads. I rebased onto it and kept its structure. Both of its encodes now take `encoderOptions`, and the no-alpha branch is what closes #1219: with no intermediate to break the lineage, the encoder still held the read-time palette and quantised the corrected pixels straight back onto their originals, so corrections had no visible effect at all.

Four chains keep a bare `toBuffer()`, each checked rather than assumed: three read an SVG, where Sharp already falls back to PNG, and meme-generator's is the answer itself, whose detected format sets the response content type.

## Tests

A generated guard over the image catalogue, as #1190 asked for. Every image tool that takes one image and gives one back runs over the same pixels twice, once as a GIF and once as the PNG transcode of that same GIF. A second sweep runs a JPEG against its own PNG transcode, exact for the tools that answer in PNG either way. A coverage case makes a new image tool fail until someone covers it or writes down why not, pinned against `REGISTRY_EXEMPT` so a tool that gains a single-buffer contract cannot keep its exemption quietly.

Settings are chosen per tool rather than taken from `defaultSettingsFor`, which matters more than it sounds: `border` defaults to a black border, and black is roughly what the corruption produces, so the tool that started this issue passes the guard on its own defaults.

The pixel guard cannot drive a tool that takes several files or shells out to a model, and those hold more than half the call sites here, so a second test reads the source and fails on any `sharp(...).toBuffer()` chain that names no format, across all of `apps/api/src`. `auto-orient` is listed there as known-unfixed against #1220 rather than as an allowance, because it is this bug and PNG is the wrong answer for it.

## What I ran

`pnpm lint` and `pnpm typecheck` clean. `pnpm test` on this branch against the same suite on an unmodified worktree at the base commit. Both runs failed exactly one test, a different SCIM race each time, which is #1156's documented shape. I checked rather than labelled it: the same file alone on this tree failed `Users PUT concurrent renames`, and the same file alone on the unmodified baseline failed the same test with the same `outcome: 403,200` licence-gate contender. Zero new failures.

Re-ran after rebasing onto #1225.

Then I reverted each piece of the fix one at a time to see which cases go red. That pass paid for itself three times:

- My first attempt at the alpha rejoin dropped transparency on the PNG side too, and the test stayed green because it only compared GIF against PNG. Both assertions are absolute now.
- The JPEG sweep could not see a symmetric lossy intermediate: swap circle-crop's `.png()` for `.jpeg({quality: 80})` and both runs pay it equally, so the difference cancels. There is now an absolute case that reads a pixel back out of the source.
- The source scanner skipped any chain with a comment inside it, and one of my own comments had made image-pad invisible to it.

## Tradeoffs

A fresh palette is not free, and the cost falls on the tools that needed it least. Reuse hands an unchanged photographic animation back byte for byte; regenerating moved a channel by up to 36 on `animated-simpsons.gif` and added 5% to the file, for about a fifth more encode time. The alternative is deciding per tool whether its settings can introduce a colour, and that judgement goes stale exactly where it matters: `border` looks like a pure frame until someone sets a colour.

PNG intermediates are larger than the JPEG ones they replace, roughly 6x on a photographic 2000x1500. `stitch` pays one per input. Raw would be worse, not better, at about 12MB against PNG's 7.8MB for that size, so the real answer is not round-tripping to a buffer at all, which is a bigger change than this.

## Filed, not fixed

#1220 auto-orient, #1221 gif-tools reverse and rotate collapsing every frame onto one palette, #1222 the same palette rule missing in `packages/image-engine`, #1223 the optimize colour count having four effective settings, #1224 a failed deep-enhance pass returning 200 in silence.


snapotter-hq:
Ran the pr-review-toolkit agents (silent-failure-hunter, code-reviewer, pr-test-analyzer) over the diff before opening this. They found six real defects, three of them in my own tests, and every one is fixed in the branch as pushed.

**In the tests, which is the embarrassing half.**

The source scanner broke off at the first comment inside a chain, so any chain carrying an explanation went unscanned. One of my own comments had done exactly that to `image-pad`: delete its `.png()` and the scanner stayed green. Comments now count as trivia, and I re-ran the deletion to watch it fail.

The scanner only read `routes/tools`. `lib/bg-effects.ts` had five live instances of this bug, reached by blur-background, remove-background and background-replace, with the same 76,105,113 signature. Widened to all of `apps/api/src`, and those five are fixed here.

`nothing is exempt that the guard could have run` was vacuous. It asked `getToolConfig`, which returns undefined for any module the test never imports, so it held even for `split`, which is registered. It now checks against `REGISTRY_EXEMPT`, which I pulled into `tests/helpers` so the drift guard and this one read the same list.

**In the fix.**

`image-enhancement`'s first encode is an intermediate when the alpha rejoin follows, so it was encoding to the output format twice: a transparent WebP came out 26 levels off on a channel. It names PNG now, which also hands `joinChannel` the three channels it needs and let me drop the extra `removeAlpha` round trip entirely.

The JPEG sweep could not see a lossy intermediate that both runs pay equally. Swap circle-crop's `.png()` for `.jpeg({quality: 80})` and the difference cancels to zero. There is now an absolute case that reads a pixel back out of the source, and I checked it fails for both that swap and the original unnamed form.

My `reuse: false` doc comment claimed the cost was "a unit or two". On `animated-simpsons.gif` it is 36 on a channel and 5% on the file, while reuse hands the same animation back byte for byte. Corrected, and so is the comment in `rotate.ts` that my change had made false.

**Considered and dismissed.**

One reviewer measured `reuse: false` adding 13 to 18 drift on still GIFs. I could not reproduce that: on a flat GIF, a noise GIF and `sample.gif` I get 2 at worst, and `reuse: true` gives the same 2. The real cost only shows on animated photographic GIFs, which is the number I documented instead.

Same reviewer read the strict JPEG case as adding nothing the GIF sweep lacks. It is the only thing covering circle-crop and rounded-crop, which answer in PNG whichever container they were handed and so are invisible to the GIF sweep. Reverting either one turns only that case red.

Several `SETTINGS` entries were flagged as not exercising palette reuse, which is true: composite drops the palette, so text-overlay, watermark-text, meme-generator and replace-color cannot show it. They stay as they are, pinning that output does not move with the container, which is what the docstring already claims for them.

Suggestion to assert `NOT_RUNNABLE` against the registry too: not taken. Those four are excluded by this environment, not by a contract, so there is nothing authoritative to check them against. They do get a staleness check now, so an entry for a tool that left the catalogue fails.

**Filed rather than fixed:** #1220, #1221, #1222, #1223, #1224. #1221 is the one worth looking at soon: gif-tools reverse and rotate render every frame in one frame's colours, and red/green/blue comes back as blue/blue/blue.

PR #1229, opened 2026-09-20
fix(image): preserve format and quality when auto-orienting

### LINKED ISSUE #1220: fix(upload): auto-orient re-encodes rotated photos at Sharp's default quality

drakeo338:
## What does this PR do?

Fixes #1220

`autoOrient` rotates images that carry an EXIF orientation tag before any tool runs. The rotate branch was `sharp(buffer).rotate().toBuffer()`, and because `toBuffer()` names no format, Sharp writes the pixels back in whatever container it read them from, at that encoder's own default quality. For a JPEG that means a second lossy generation at roughly quality 80, dropped in ahead of the tool the user actually asked for. Phone photos set an orientation tag almost every time, so in practice this was quietly re-compressing most camera-roll uploads, while an already-upright photo (orientation 1) skipped the branch and was untouched.

The fix is to name the format and the quality instead of letting Sharp guess. I route through `resolveOutputFormat`, the same helper the tool routes already use, and encode with `.toFormat(output.format, output.encoderOptions)`. That gives quality 95 to match the rest of the pipeline, and it also picks up the GIF palette rule for free rather than re-deriving it here.

The part I went back and forth on was whether to just force PNG. The related #1190 cleanup settled most of these unnamed-intermediate sites by adding `.png()`, and doing the same here would have been the smallest diff. But `autoOrient` sits on most upload paths and hands its output straight to the tool, so turning every rotated JPEG into a PNG would blow up what the rest of the pipeline carries. Preserving the input container is the right call here: JPEG stays JPEG, PNG stays PNG. Going through `resolveOutputFormat` also means a format with no Sharp encoder (camera RAW, SVG and friends) falls back to PNG, which is the only thing you can do when the container can't be written back anyway; none of those carry EXIF orientation today, so the branch never reaches them, but I didn't want the fix to depend on that staying true.

On the tests: `tests/unit/api/auto-orient.test.ts` already builds synthetic images in memory, so I added a case there that rotates a JPEG and checks the result is still a JPEG and matches a quality-95 re-encode rather than Sharp's default. The image is filled with deterministic noise on purpose, because a flat fill compresses to the same size at any quality and the assertion would pass either way. I checked that this case fails against the old `toBuffer()` (10114 bytes at the default vs 15190 at quality 95) so it genuinely guards the regression. I also removed the `lib/auto-orient.ts` entry from `KNOWN_UNFIXED` in `tests/unit/api/sharp-intermediate-formats.test.ts`; with it gone, that file's source scan enforces the fix and would fail again if this site ever regressed.

## Type of change

## Checklist

- [x] The files this PR touches pass locally. For full disclosure, `pnpm test` has 8 pre-existing failures in `tests/unit/ai/runtime-state.test.ts` that also fail on a clean `main` and are unrelated to this change (image handling).


drakeo338:

drakeo338:
recheck

drakeo338:
recheck

Issue #1243, opened 2026-09-21
test: the integration test server skips the production error handler, so thrown-error response bodies are untested


snapotter-hq:
## What happens

`buildTestApp` in `tests/integration/test-server.ts` mirrors `apps/api/src/index.ts` plugin by plugin (upload, cookie, auth, tool access, rate limit, every route group) but never installs the production error handler. Any error thrown out of a route in an integration test is rendered by Fastify's default handler as `{ statusCode, error: "Service Unavailable", message }`, while production renders `{ error, details?, code? }` through `registerErrorHandler` (`apps/api/src/plugins/error-handler.ts`, extracted from `index.ts` in the PR for #1161).

So no integration test can assert what a client sees when a route throws. The batch ingress test added for #1161 (`tests/integration/platform/batch.test.ts`, "deletes the files it already stored when a later file cannot be stored") had to assert `res.body` contains the message rather than the production shape, and the production rendering is pinned only by the unit test `tests/unit/api/error-handler.test.ts`.

## Where it lives

- `tests/integration/test-server.ts`, `buildTestApp`, right after `Fastify({...})`: no `registerErrorHandler(app)` call.
- `apps/api/src/plugins/error-handler.ts`: the handler, now importable.
- `tests/integration/platform/posthog-proxy.test.ts` around line 264 already notes the production handler treats every 5xx as a bug and installs its own stub, which suggests other tests have adapted to the gap.

## Why it was out of scope

Wiring it in changes the response body for every thrown error across roughly 290 integration files that assert on status codes and 30 that assert on a `.message` field. Some of those assertions match Fastify's default shape and would need rewriting; #1161 was not the place to sweep them.

## Starting sketch

Add `registerErrorHandler(app)` to `buildTestApp` in the same position `index.ts` uses, run `pnpm test:integration`, and fix each assertion that was matching the default shape (`body.message`, `body.error === "Bad Request"`, `body.statusCode`). Then tighten the #1161 ingress test to assert `{ error, code: "workspace-cap" }` directly.

## Gotchas

- The `posthog-proxy` test's own `setErrorHandler` stub must keep winning inside its scope, or its 5xx expectations change.
- Fastify validation errors (`FST_ERR_VALIDATION`) render differently under the two handlers; check the OpenAPI examples in `apps/api/src/openapi.yaml` still describe the production shape.

Issue #1242, opened 2026-09-21
fix(web): single-file and pipeline runs only degrade on 502 and 504, so a Cloudflare 524 still fails them


snapotter-hq:
## What happens

The PR for #1161 taught the batch path to treat any 5xx with a non-JSON body as an intermediary answering for a dead sync wait, and to ride the SSE instead of failing. The single-file path and both pipeline paths still only do that for 502 and 504, so a Cloudflare 524 (its 100 s origin limit), a 520/522/523, or a load balancer's HTML 503 still fails the run while the server keeps processing.

Repro: put Cloudflare, or any proxy that answers its own timeout with a non-502/504 status, in front of an instance with `SYNC_WAIT_MS` raised past the proxy's limit (or a pipeline batch, which holds its response longer). Run a single `compress`, or execute a pipeline. Expected: the run degrades to the SSE path and settles from the terminal frame. Actual: "Processing failed: 524" and a job that completes server-side with no client watching it.

Low priority for single-file runs: the default `SYNC_WAIT_MS` is 8 s, so a 100 s proxy limit never fires unless an operator raises it. Pipelines are the more likely victim.

## Where it lives

- `apps/web/src/hooks/use-tool-processor.ts`, the single-file `xhr.onload` handler: `if (xhr.status === 502 || xhr.status === 504)` with the `appSpoke` JSON probe.
- `apps/web/src/hooks/use-pipeline-processor.ts`: the same 502/504 pair in both `degradeToAsync` call sites (around lines 498 and 880) and the narrow `trigger` union on its `trackDegrade`.
- The batch path's rule to copy is in `use-tool-processor.ts` after #1161: `xhr.status >= 500 && degradeToAsync(\`http-${xhr.status}\`)` inside the non-JSON branch. `ToolRunDegradedProperties.trigger` in `packages/shared/src/analytics/events.ts` already accepts `http-<status>`.

## Why it was out of scope

#1161 was about batch failures; widening the other two paths is the same three-line change but in functions the issue did not touch, and each has its own recovery test file to extend (`tests/unit/web/use-tool-processor-single-file-failure.test.ts`, `tests/unit/web/use-pipeline-processor-recovery.test.ts`).

## Starting sketch

[...]

Issue #1241, opened 2026-09-21
fix(api): one large batch keeps the workspace full for FILE_MAX_AGE_HOURS and fails every later request


snapotter-hq:
## What happens

One large batch can fill the workspace and keep it full for three days. Inputs, outputs and the result ZIP all live under `MAX_WORKSPACE_SIZE_GB` (10 GB by default) and nothing frees them until the TTL sweep, `FILE_MAX_AGE_HOURS` (72 by default). A batch whose inputs plus outputs cross the cap fails at packaging, its child outputs are unreachable (no ZIP, and the UI never links child jobs), and every request on the instance after it, single-file or batch, gets the cap's 503 until the sweep runs.

Sentry NODE-5Y has the pattern on five instances. The clearest: a 2992-file `sharpening` batch on 2026-09-20 completed 2753 children, failed, and the user's next eight 8-file retries all 503'd within two minutes. NODE-6Z is the same cap hitting single-file `heic-to-jpg` on another instance.

Repro: set `MAX_WORKSPACE_SIZE_GB=1`, run a 20-file batch of 30 MB images through `convert`. Expected: the batch fails with a reason and the instance keeps working. Actual (after the PR for #1161): the batch fails with the cap's message and the batch's own uploads are cleaned up, but the child outputs stay for 72 hours and the next request still fails.

## Where it lives

- `apps/api/src/jobs/worker.ts`, `processBatchFinalize`: on a packaging failure it records the terminal state and rethrows; it never deletes `uploads/<parent>-f*` or `outputs/<parent>-f*`.
- `apps/api/src/jobs/worker.ts`, `processBatchChild` / `processToolJob`: a child's input under `uploads/` is never needed again after it completes, but stays until the sweep.
- `apps/api/src/jobs/system-jobs.ts`, the storage TTL sweep (around line 290): age-based only, runs every `CLEANUP_INTERVAL_MINUTES`.
- `apps/api/src/lib/object-storage.ts`, `assertWorkspaceSizeCap`: refuses the write; it does not try to reclaim anything first.

## Why it was out of scope

The PR for #1161 refuses batches that cannot fit up front and cleans up a batch's uploads when ingress fails. Reclaiming space from completed or failed jobs is a retention policy change: it decides what a user can still download later, so it needs the product call, not a side effect of a bug fix.

## Starting sketch

Three independent pieces, any of which helps:

[...]

Issue #1240, opened 2026-09-21
fix(web): batches over about 100 files fail because the client unzips the whole result archive in memory


snapotter-hq:
## What happens

Batches of 100 or more files fail most of the time. PostHog, 60 days on 2.2.0: 22 batches of 100+ files, 17 of them `batch_processed` with `status: failed`, across 11 deployments. The largest, 2992 files through `sharpening`, had 2753 children complete server-side (`tool_used` events) before the run failed on the client 92 s after the last child.

The client settles a batch by pulling the whole ZIP into memory: `settleFromZip` calls `zipBlob.arrayBuffer()` and then fflate's `unzipSync` on it, so a result archive needs about twice its own size in browser memory, and Chromium caps a single ArrayBuffer well below what a multi-GB ZIP needs. The failure is a RangeError inside `settleFromZip`, caught and reported as "Batch processing failed" with no hint that the files are fine and sitting in a ZIP the server still has.

Repro: run any fast image tool over 1,000 files of 3 MB. Expected: results come back, or the UI offers the ZIP as a download. Actual: every child completes, the ZIP is written, the client shows "Batch processing failed" and the outputs are unreachable from the UI.

## Where it lives

- `apps/web/src/hooks/use-tool-processor.ts`, `settleFromZip` (around line 917): `arrayBuffer()` plus `unzipSync`, then one `URL.createObjectURL` blob per entry, so the extracted files are held a third time as blobs.
- The sync path hands it `xhr.response` (a Blob of the whole ZIP, `xhr.responseType = "blob"`); the degraded path downloads the same ZIP with `fetch(...).blob()` in `downloadAndSettle`.
- After the PR for #1161, `batch_processed` carries `reason`; this path reports `unzip-failed`, and a failed ZIP download reports `download-failed`, so the split between the two will be visible in analytics once released.

## Why it was out of scope

#1161 was about batches failing without a reason. This is a size ceiling in the result path and needs a design decision (stream the ZIP, or stop extracting above a threshold).

## Starting sketch

The terminal frame and the sync response both know the ZIP size before the client touches it (`processedSize` in `result`, `Content-Length` on the response). Above a threshold, skip extraction: keep the ZIP as the batch result, show one download button, and mark the entries as completed without per-file previews. Below it, keep today's behaviour. fflate's streaming `Unzip` would let the extraction run entry by entry without the full-archive buffer, but it still leaves one blob per entry in memory, so the threshold is needed either way.

## Gotchas

[...]

PR #1245, opened 2026-09-21
fix(batch): refuse uploads that cannot fit and ride out proxy 5xx responses

### LINKED ISSUE #945: fix(web): degraded-batch evidence timeout never emits the batch_processed analytics event
Every batch outcome emits BATCH_PROCESSED via trackBatch (completed, failed, canceled through failRun and settleFromZip), except one: a degraded batch (#750 socket-death path) whose job-evidence timeout fires. The timeout callback in apps/web/src/hooks/use-tool-processor.ts (the JOB_EVIDENCE_TIMEOUT_MS teardown, ~lines 200-225 after #943) nulls batchRunRef and settles the run directly, so failRun never runs and the event is never sent. Operators watching batch failure rates in PostHog miss exactly the runs a flaky proxy kills, which is the population #750 exists for.

Repro: batch run, upload completes, POST socket dies (degrade), no SSE frame ever arrives, 30s evidence timeout fires. UI settles correctly (after #943 the entries fail too), but no batch_processed event lands.

Out of scope of #943 (analytics, not the stuck-pulse settle). Fix sketch: the timeout callback knows it ended a batch when batchRunRef was non-null on entry; emit the same event failRun sends with status "failed" before nulling it. Gotcha: trackBatch lives inside processAllFiles's closure, so either hoist the emit or capture what it needs on the ref. Remember the TELEMETRY.md rule if the event shape changes (it should not; reuse batch_processed as-is).

Found during the silent-failure review of PR #943; pre-existing, unchanged by that PR.

### LINKED ISSUE #1161: Batch runs fail at 23%, and failure scales with file count (34 files average when failed, 11 when completed)
Filed from analytics, not from a repro. I could not reproduce it locally and the mechanism below is a shortlist, not a diagnosis. The correlation is strong enough to be worth someone's afternoon.

## The signal

60 days of `batch_processed` on 2.2.0 (the event fires once per batch as a unit, `apps/web/src/hooks/use-tool-processor.ts:801`):

| status | batches | instances | mean `file_count` |
| --- | --- | --- | --- |
| completed | 505 | 167 | **10.9** |
| failed | 151 | 56 | **34.2** |

23% of batches fail, and failed batches carry roughly three times as many files as successful ones. Batch size is the strongest predictor of failure we have on any subsystem.

56 distinct deployments hit this, so it is not one person with a bad setup.

## What I ruled out

- Not a single tool. The failures spread across tools rather than concentrating.
- Not `MAX_BATCH_SIZE`. It defaults to `0` (unlimited) in `.env.example:14`, and a config rejection would surface as a clean error at submit, not a failed run.
- Not the QA sweep instances; they are excluded from these numbers.

## Where to look

Ordered by how much I believe them:

1. **ZIP packaging timeout.** `BATCH_ZIP_TIMEOUT_MS` is 15 minutes (`apps/api/src/jobs/worker.ts:1311`), and the archiver pass at `worker.ts:1317-1366` streams every child output through one `archiver` instance. Thirty-four files of any real size on a CPU-only box is plausibly past that. This is my first guess because the failure would land exactly where the data points: at the end, after the work, scaling with count.
2. **Per-user in-flight caps.** `.env.example:17` caps concurrent single-file AI jobs per user. If a large batch trips that, check whether the rejection settles the batch as failed rather than queueing.
3. **Memory during finalize.** Worth measuring peak RSS on a 40-file batch before assuming it is time and not space, given how much of the fleet is CPU-only with modest RAM.

## How to reproduce

Run `compress` or `convert` over 40 files on a CPU-only container with a 2 GB limit, which is the low end of what the fleet actually runs. Watch whether the finalize job or the archiver pass is what gives out. If neither does, bisect on file size rather than count, since `file_count` is the only dimension the event carries and total bytes may be the real variable.

Instrumenting that is probably step one: `batch_processed` has `file_count` and `status` and nothing else. Adding total input bytes and a failure reason would turn this issue from a correlation into an answer, and it is a small change next to the fix.

## Prior art

- #767 and #770 built the batch cancel path and the partial-ZIP semantics; `processBatchFinalize` and its canceled-children counting are the same code this would touch.
- #951 (open) notes batch child failures never reach Sentry, which is very likely why this has been invisible until now. Fixing that first might just answer this issue outright.
- `tests/integration/platform/batch-cancel.test.ts` has a worker harness with real flows that a size-scaling test could reuse.

## Acceptance

Either a 40-file batch completes on a 2 GB CPU-only container, or it fails with a reason the user can act on ("batch too large for available memory, try fewer files"). The current outcome is a generic failure after several minutes of work, which is the worst of both.

snapotter-hq:
Closes #1161
Closes #945

## What the data said

#1161 came from a correlation (23% of batches fail, failed batches carry three times the files) and three guesses, all at the end of the run: ZIP timeout, AI job caps, memory during finalize. None of the three held. Joining `batch_processed` with the server-side `tool_used` events (both carry the browser's distinct id) and with Sentry gave a different picture:

- The released 2.2.0 client predates every batch resilience fix on main (#751, #765, #769, #770, #798, #931, #943). On 2.2.0 any dead sync response fails the batch and the route answered 422 after its 30-minute wait. Failed runs cluster at 58 to 62 s (13 failed vs 6 completed, nginx's `proxy_read_timeout` default) and 95 to 105 s (10 vs 7, Cloudflare's 524). Most of that is already fixed on main and just unreleased.
- Sentry NODE-5Y: the workspace cap (`MAX_WORKSPACE_SIZE_GB`, 10 GB default) tripping inside the batch route on five instances, 20 events, each timestamp matching a failed `batch_processed` row. A 2992-file batch completed 2753 children, failed, and the user's next eight small retries all failed within two minutes because the workspace stayed full. The client showed "Internal server error".
- 22 batches of 100+ files in 60 days, 17 failed. Filed as #1240.

What was still broken on main and is fixed here: the cap path and the non-502/504 proxy 5xx.

## What changed

Server:

- `apps/api/src/routes/batch.ts` compares `Content-Length` with the workspace headroom before buffering the body and answers 503 with the numbers and the two knobs (`MAX_WORKSPACE_SIZE_GB`, `FILE_MAX_AGE_HOURS`). Chunked uploads have no length and fall through to the per-write check.
- The same route tracks every key it stores, not only the OCR ones, and deletes them if ingress fails before the flow is enqueued. That is what kept a full workspace full. It also settles the parent row, drops the staged child rows, and publishes the terminal frame, so job history and an SSE-only client see the reason instead of "processing".
- When the sync wait rejects because the finalize failed, the route answers from the row the finalize settled (message, code, per-file errors; 503 for a capacity code). BullMQ's rejection is a plain Error the handler would otherwise mask.
- When every child failed for the same reason, that reason is the batch's error on the 422 body and the frame instead of "All files failed processing".
- `apps/api/src/lib/object-storage.ts` exports `workspaceHeadroomBytes()` sharing the existing 30 s size cache, drops that cache on every delete, and the cap's message no longer says "try again shortly" (it was three days).
- The error handler moved from `index.ts` into `apps/api/src/plugins/error-handler.ts` and passes a `SafeError`'s message and `code` through when the error was authored with an HTTP status. The AI bridge's SafeErrors carry sidecar output and no status, so they keep the mask.
- `processBatchFinalize` keeps a `SafeError`'s message (through `friendlyError`) and code instead of "Failed to package batch results".

Client (`apps/web/src/hooks/use-tool-processor.ts`):

- A non-JSON 5xx after the upload finished degrades the batch to the SSE path whatever the status, not only 502 and 504. Cloudflare's 524 was failing runs the server went on to complete.
- `batch_processed` gains `reason` (the server's `code` when the body has one, else `http-<status>`, `socket`, `all-files-failed`, `unconfirmed`, ...; `canceled` on canceled runs) and `total_bytes`, so the next analysis is a diagnosis rather than a correlation. `TELEMETRY.md` documents both.
- The job-evidence teardown now ends a batch through its own failure path, so abandoned runs appear in `batch_processed` (#945). A stale batch closure refuses the call, like `cancelLocally`, so a later single run still settles.
- `trackBatch` uses the static `track` import the hook already had; the dynamic import bought nothing and hid the event from fake-timer tests.

## What I ran

Every new test was watched failing first: the first commit's cases were reverted against once more as a whole (9 unit cases, the handler file, 3 integration cases all fail without the fix), and the second commit's nine cases were run red before the code.

- `pnpm vitest run tests/unit/api tests/unit/shared`: 221 files, 3825 passed.
- `pnpm vitest run tests/unit/web`: 137 files, 1955 passed (every consumer of the hook), plus the five touched files again after the second commit: 81 passed.
- `pnpm vitest run tests/integration/platform/batch.test.ts`: 54 passed.
- `pnpm lint`, `pnpm typecheck`: clean.
- Batch e2e specs (`gui-batch`, `batch-preview`, `gui-multi-file-workflows`, chromium, one worker): 66 passed, run once per commit (3.9 min and 2.9 min).
- Full `pnpm test` before and after (`VITEST_MAX_FORKS=2`, about 42 min each), diffed on the failing-test lines: identical. Baseline 893 files and 20562 tests passed with one failure; this branch 894 files and 20584 tests passed with the same one failure, the SCIM concurrent-create race (#1156, `outcome: 201,403` in both runs). Nothing new, nothing gone.

## Left as is, on purpose

- The batch route still holds the HTTP response for up to 30 minutes while the docs advertise a synchronous ZIP in 20 locales. Returning 202 after `SYNC_WAIT_MS` like single-file tools would sidestep every proxy timeout at once, but it changes a documented contract, so it is a suggestion here rather than a change.
- Single-file and pipeline runs keep their 502/504-only rule (#1242).
- The integration test server runs Fastify's default error handler, so the ingress cleanup test asserts on the message text rather than the production body shape; the production shape is pinned by the new unit test (#1243).
- Ingress still buffers the whole upload in memory (#1239), and a failed large batch's child outputs stay for `FILE_MAX_AGE_HOURS` (#1241).
- A partial batch (some children hit the cap, the rest succeed) still labels the missing entries "File not found in batch results"; the per-file reasons are in the frame's `errors` but the client does not map them onto entries yet. Noted in #1241.


snapotter-hq:
Pre-PR review with the pr-review-toolkit agents (silent-failure hunter, code reviewer, test analyzer) on the first commit. What each found and what happened to it:

**Fixed in the second commit**

- Sync path still masked (hunter, analyzer, both rated it the top finding). The finalize wrote the cap message to the row and the frame, but BullMQ rejects the route's wait with a plain Error, so the browser on the sync path and curl callers still saw "Internal server error". The route now answers from the settled row.
- Parent and staged child rows left at "queued" after an ingress failure (hunter, analyzer). Settled and dropped in the route's catch, with the terminal frame published.
- Every child failing on the cap surfaced as "All files failed processing" (hunter). The shared reason is now the batch's message on the 422 and the frame.
- The 30 s size cache kept saying "full" for up to 30 s after the cleanup freed space (reviewer). Deletes drop the cache.
- The SafeError pass-through cited an invariant the AI bridge breaks (reviewer). Gated on a status-bearing SafeError; the finalize runs the message through `friendlyError`.
- `abandon` skipped the run-identity guard (hunter). Added, with a test where a single run starts on top of a degraded batch.
- `reason` collapsed the cap into `http-503` (hunter). The body's `code` wins when present.
- Canceled runs reported `server-failed` (analyzer). They report `canceled`.
- Pins the analyzer asked for: the real cap error's code and message, the S3 headroom branch, three handler branches, a JSON 5xx on the batch path that must not degrade, and the stale "dynamic import" comment.
- Cleanup failures now log at warn (hunter). Two comment pointers to the old handler location updated (reviewer).

**Considered and left**

- Degrading on any non-JSON 5xx turns a proxy's own 503 (backend down) into a 30 s evidence wait instead of an immediate failure (hunter). Kept: the wait ends with "the server never confirmed the job", which is the accurate story, and #750 accepted the same trade for 502.
- Per-file cap errors on a partial batch are not mapped onto entries (hunter). Client work with its own design; noted on #1241.
- Chunked uploads skip the pre-flight and are untested (analyzer). Documented in the code; the per-write check still applies.
- Wiring the production error handler into the integration test server (analyzer, reviewer). Ripples through files that assert Fastify's default shape; #1243.
- Pre-flight on inputs only, not inputs plus outputs plus ZIP (reviewer). A multiplier would be a guess; the packaging path now carries the message.
- `/api/v1/pipeline/batch` and single-file uploads have no pre-flight (reviewer). Out of this issue; the same helper applies when someone picks it up.

snapotter-hq:
Rebased onto 76a1de52 (#1244) before merging and re-ran the full suite on the rebased head: 894 files, 20587 tests, one failure, the SCIM concurrent-rename race from #1156 (a different case than the baseline's, same 403 contender from the licence gate; the unfixed tree fails the same file the same way under the same load). CI on the rebased head: the landing e2e lane timed out waiting 120 s for the Astro build to serve, and integration shard 2 hit the same SCIM race; both green on re-run, nothing in this diff touches either.

